PR #9, opened 2018-02-02
Launcher script for Argo workflows to be used by prow jobs.

jlewi:
* This is a script from kubeflow/kubeflow used by our prow jobs to trigger
  and wait for an Argo workflow.
   * Pull in a bunch of helper scripts as well; e.g. the Argo client.

* We generalize the script so it can be used to launch different workflows for different repositories.


README.md
```diff
@@ -14,10 +24,17 @@ The current thinking is this will work as follows
...
-The UI is publicly available at http://http://testing-argo.kubeflow.io/
+* Our prow jobs are defined in [config.yaml](https://github.com/kubernetes/test-infra/blob/master/prow/config.yaml)
+* Each prow job defines a K8s PodSpec indicating a command to run
+* Our prow jobs use [run_e2e_workflow.py](https://github.com/py/kubeflow/testing/run_e2e_workflow.py
+) to trigger an Airflow pipeline that checks out our code and runs our Tests.
+* Our tests are structured as Argo pipelines so that we can easily perform steps in parallel.
+
+## Accessing The Argo UI
 
+The UI is publicly available at http://http://testing-argo.kubeflow.io/
```
lluunn: typo in the url

images/Dockerfile
```diff
@@ -1,24 +1,12 @@
 # Docker image for running E2E tests using Argo.
 
-FROM python:2.7-slim
+FROM ubuntu:xenial
```
lluunn: just curious: why do we change to xenial?

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -0,0 +1,230 @@
+"""Run the E2E workflow.
```
lluunn: why is the directory structured as py/kubeflow/testing/xxx? Isn't "kubeflow/testing" part unnecessary?

images/Dockerfile
jlewi: python2.7 was a vestige of the Airflow image we were originally. Using.
It makes more sense to use a generic linux image.

py/kubeflow/testing/run_e2e_workflow.py
jlewi: We want this to be a python package that is imported as "kubeflow.testing".
The idea here is that we can add "py" to our python path and then import the code as
import kubeflow.testing...

jlewi:
Thanks. PTAL.

lluunn:
LGTM.

PR #17, opened 2018-02-09
Add argo workflow status link

jose5918:
Reasons for change:
- Would like a link directly from a PR to the corresponding argo workflow to make debugging easier

Changes made:
- Add calls to run_e2e_workflow.py for setting github status to pending, success, or failure
- Adds github_status.py for sending status update


jlewi:
We'll need to rebuild the Docker images for this to take effect.

Issue #22, opened 2018-02-10
Add argo status link to GitHub


jlewi:
Would like a link directly from a PR to the corresponding argo workflow to make debugging easier

This was originally aded in #17 but this broke things #20 and was rolled back.

The environment variable GIT_TOKEN isn't defined in the prow job. This code runs in the prow container.

We'll need to rethink how we get the GIT API token; its possible prow attaches a secret already for the API token.

Alternatively it might just be easier to see if we can set it in gubernator see #10 .

/cc @jose5918

gaocegege:
FYI, I filed [an issue](https://github.com/argoproj/argo/issues/733) to the upstream and they recommend a new project: [argo-ci](https://github.com/argoproj/argo-ci), which is built on argo for CI.

jlewi:
What would argo-ci give us over prow?

gaocegege:
@jlewi They said argo-ci supports status link internally compared to argo.

jlewi:
If we use our build cluster with prow (#32) we can easily add whatever secrets we need.

@gaocegege I think argo-ci provides overlapping functionality with the dashboards and reporting in prow; but prow is more mature and battle tested by Kubernetes.

PR #28, opened 2018-02-12
Argo status

jose5918:
Reasons for change:
- Would like a link directly from a PR to the corresponding argo workflow to make debugging easier

Changes made:
- Add calls to run_e2e_workflow.py for setting github status to pending, success, or failure
- Adds github_status.py for sending status update
- Gets github token from `kubernetes-test-infra` namespace

https://github.com/kubeflow/testing/issues/22

### DIFF
--- a/py/kubeflow/testing/github_status.py
+++ b/py/kubeflow/testing/github_status.py
@@ -1,0 +1,81 @@
+#!/usr/bin/python
+"""Post status back from an argo-workflow
+
+Requires Prow environment variables to post back status
+"""
+
+import argparse
+import json
+import logging
+import os
+import requests
+
+# Repo org and name can be set via environment variables when running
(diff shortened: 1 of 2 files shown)


jose5918:
@jlewi Is there a way to test this?

jose5918:
WIP now since there are conflicts and needs support for many workflow, so I will probably just list urls in gubernator

jlewi:
@jose5918 If you wanted to test this we could setup a prow job for this repo (this is something we want to do). You could copy the workflow used by tf-operator and just run the py lint and unittest steps.

The presubmits would then use the code in this PR to run the presubmit which would be a good way to verify nothing breaks.

jlewi:
@jose5918 should we close this PR?

PR #34, opened 2018-02-15
add pyyaml, pipfile and lock requirement.txt with hash

Jimexist:
add pyyaml, pipfile and lock requirement.txt with hash


images/Pipfile
```diff
@@ -0,0 +1,54 @@
+[[source]]
+
+url = "https://pypi.python.org/simple"
```
jlewi: Is the Docker file actually using this pip file?

Jimexist: no, but i think it is easier for later changes. i can try to make sure they are in sync or better tho we should use `pipenv` instead. WDYT

jlewi: SGTM. I'll merge this as is and we can figure it out in a follow on PR.

PR #38, opened 2018-02-21
Taking junit results into account when computing prow job status.

### LINKED ISSUE #27: Need to take test status into account when reporting prow job success or failure
We need to look at the junit XML files to decide whether to report the job as failed or success

* Right now we just look at whether the Argo workflow succeeded or not which is insufficient since tests can fail but Argo workflow can still succeed.

We need to do something like tensorflow/k8s#319 

If we want to be strict about which xml files are expected we could specify the xml files expected in the prow_config.yaml file.

jlewi:
* When decided whether to mark a prow job as failed or succeded we need
  to explicitly look at the junit files and determine that.

* Prow won't do this automatically.

* Fix #27

### DIFF
--- a/py/kubeflow/testing/prow_artifacts.py
+++ b/py/kubeflow/testing/prow_artifacts.py
@@ -9,6 +9,7 @@ import json
 import os
 import time
 from google.cloud import storage  # pylint: disable=no-name-in-module
+from kubeflow.testing import test_util
 from kubeflow.testing import util
 
 # TODO(jlewi): Replace create_finished in tensorflow/k8s/py/prow.py with this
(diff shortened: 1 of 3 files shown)


py/kubeflow/testing/prow_artifacts.py
```diff
@@ -147,6 +148,66 @@ def create_pr_symlink(args):
...
+
+  output_dir = get_gcs_dir(bucket)
+  artifacts_dir = os.path.join(output_dir, "artifacts")
+
+  # If the workflow failed then we will mark the prow job as failed.
+  # We don't need to check the junit files for test failures because we
+  # already know it failed; furthermore we can't rely on the junit files
+  # if the workflow didn't succeed because not all junit files might be there.
+  if workflow_success:
+    workflow_success = check_no_errors(gcs_client, artifacts_dir, junit_files)
```
lluunn: Test failed: global name 'junit_files' is not defined

jlewi:
@lluunn Tests are passing.

PR #39, opened 2018-02-21
use prow_config if exists

lluunn:
Check prow_config.yaml exists before loading it.

We will add config_file arg for kubeflow/kubeflow in https://github.com/kubeflow/kubeflow/pull/275

### DIFF
--- a/py/kubeflow/testing/run_e2e_workflow.py
+++ b/py/kubeflow/testing/run_e2e_workflow.py
@@ -103,9 +103,9 @@ def run(args, file_handler): # pylint: disable=too-many-statements
   workflows = []
   if args.config_file and os.path.isfile(args.config_file):
     workflows.extend(parse_config_file(args.config_file, args.repos_dir))
-
-  if args.app_dir and args.component:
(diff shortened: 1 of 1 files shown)


jlewi:
This won't work as expected.

I think you want to modify
https://github.com/kubeflow/kubeflow/blob/master/testing/run.sh

And then rebuild the Docker image
https://github.com/kubeflow/kubeflow/tree/master/testing

That's because our prow config isn't using the generic worker image yet.

We need to update our prow configs to do that.

jlewi:
Hang on. May have spoken two hastily.

jlewi:
I think I was correct.
See
kubeflow/kubeflow#279

lluunn:
https://github.com/kubeflow/kubeflow/pull/275 modifies run.sh.
If that looks good, I will then update the image.

lluunn:
related https://github.com/kubeflow/kubeflow/issues/276#event-1485784512

lluunn:
Update https://github.com/kubeflow/kubeflow/pull/275 as suggested. Doesn't need this change.

PR #36, opened 2018-02-15
use pipenv in image building dockerfile

Jimexist:
ready for review now

PR #48, opened 2018-02-28
Enable travis and do configurations

Jimexist:
this is cherry-picked from #36 so that it can unblock others.

@jlewi but it seems the CLA status is blocking...


Jimexist:
@jlewi seems CLA bot didn't re-check. Can you help manually merge?

PR #61, opened 2018-03-09
Change test cluster from mlkube-testing to kubeflow-ci

lluunn:
Related: #18


jlewi:
Note we will also need to rebuild the image and update individual worklows to use the kubeflow-ci project.

PR #64, opened 2018-03-15
Upgrade ksonnet from 0.8 to 0.9 for testing image

lluunn:
Needed in https://github.com/kubeflow/kubeflow/pull/426


ankushagarwal:
@lluunn Does this image needs to be built and pushed?

PR #65, opened 2018-03-15
Fix Dockerfile for testing image

lluunn:
#64 failed to update one line


ankushagarwal:
The tests are failing because - by upgrading the ksonnet version, it's API has changed. We need to run `ks upgrade` on every existing ksonnet application.

PR #70, opened 2018-03-19
Poll for subprocess output while the process is running.

jlewi:
* Don't wait for the process to finish to log the output.
* This is more convenient for streaming the output of long running commands.
* We no longer use Airflow so redirecting to a file and then reading that
  file should no longer be necessary.
* Change run to also return output and have run_and_output should be a wrapper around run
   * We can delete run_and_output after updating the callers.

### DIFF
--- a/py/kubeflow/testing/util.py
+++ b/py/kubeflow/testing/util.py
@@ -57,18 +57,22 @@ def run(command, cwd=None, env=None, dryrun=False):
     command, cwd=cwd, env=env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
 
   logging.info("Subprocess output:\n")
+  output = []
   while process.poll() is None:
     process.stdout.flush()
     for line in iter(process.stdout.readline, ''):
+      output.append(line.strip())
       logging.info(line.strip())
 
   process.stdout.flush()
   for line in iter(process.stdout.readline, ''):
+    output.append(line.strip())
     logging.info(line.strip())
 
   if process.returncode != 0:
-    raise subprocess.CalledProcessError("cmd: {0} exited with code {1}".format(
-      " ".join(cmd), process.returncode))
+    raise subprocess.CalledProcessError(process.returncode,
+                                        "cmd: {0} exited with code {1}".format(
+                                        " ".join(command), process.returncode), "\n".join(output))
 
 def run_and_output(command, cwd=None, env=None):
   logging.info("Running: %s \ncwd=%s", " ".join(command), cwd)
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/util.py
```diff
@@ -47,26 +47,32 @@ def run(command, cwd=None, env=None, dryrun=False):
...
-                            stderr=hf)
-  finally:
-    with open(log_file, "r") as hf:
-      output = hf.read()
-    logging.info("Subprocess output:\n%s", output)
+  
+  if dryrun:
+    command_str = ("Dryrun: Command:\n{0}\nCWD:\n{1}\n"
+                   "Environment:\n{2}").format(" ".join(command), cwd, env)
+    logging.info(command_str)
```
lluunn: Should dry run return here?

py/kubeflow/testing/util.py
```diff
@@ -47,26 +47,32 @@ def run(command, cwd=None, env=None, dryrun=False):
...
+
+  process.stdout.flush()
+  for line in iter(process.stdout.readline, ''):
+    output.append(line.strip())
+    logging.info(line.strip())
+
+  if process.returncode != 0:
+    raise subprocess.CalledProcessError(process.returncode,
+                                        "cmd: {0} exited with code {1}".format(
+                                        " ".join(command), process.returncode), "\n".join(output))
```
lluunn: output should be printed before, do we need to print again?

py/kubeflow/testing/util.py
```diff
@@ -47,26 +47,32 @@ def run(command, cwd=None, env=None, dryrun=False):
...
+    command_str = ("Dryrun: Command:\n{0}\nCWD:\n{1}\n"
+                   "Environment:\n{2}").format(" ".join(command), cwd, env)
+    logging.info(command_str)
+
+  process = subprocess.Popen(
+    command, cwd=cwd, env=env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+  logging.info("Subprocess output:\n")
+  output = []
+  while process.poll() is None:
```
lluunn: sleep for each loop?

py/kubeflow/testing/util.py
jlewi: Deleted dryrun. Its obsolete.

py/kubeflow/testing/util.py
jlewi: Output is being included in the exception. This makes it consistent with the exception that would be raised if we used subprocess.check_call

py/kubeflow/testing/util.py
jlewi: Done

PR #84, opened 2018-03-27
Cleanup the ${SRC_DIR}

ankushagarwal:
Since we use a shared mount, during re-runs of release,
git clone will fail. So clean up everything under the
${SRC_DIR} to start from a clean directory

Also cleanup trailing whitespaces

### DIFF
--- a/images/checkout.sh
+++ b/images/checkout.sh
@@ -2,7 +2,7 @@
 #
 # This script is used to bootstrap our prow jobs.
 # The point of this script is to check out repositories
-# at the commit corresponding to the Prow job. 
+# at the commit corresponding to the Prow job.
 #
 # In addition to understanding the prow environment variables.
 # the environment variable EXTRA_REPOS can be used to specify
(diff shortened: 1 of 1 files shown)


images/checkout.sh
```diff
@@ -16,7 +16,10 @@ set -xe
 
 SRC_DIR=$1
 
-mkdir -p /src/${REPO_OWNER}
```
lluunn: do you need to mkdir for {SRC_DIR}/${REPO_OWNER}?

ankushagarwal: Done

ankushagarwal: I looked at git clone and it seems to create any nested directories during clone by itself.

`git clone <source> a/b/c/d` would create directory structure `a/b/c/d` if not already present

jlewi:
Can we raise an error rather than rm -rf  the dir?

Instead if we rerun builds we should make the SRCDIR unique. 

I thought the releasing instructions already did this because they include a random build number
https://github.com/kubeflow/kubeflow/blob/master/releasing.md

ankushagarwal:
I see. Recreating a new build number also works.

I don't think there is a need to explicitly raise the error since `git clone` will fail if the directory already exists.

PR #85, opened 2018-03-28
Upgrade to v1.10 for kubectl

jlewi:
* v1.10 should print out the actual service account with RBAC errors
  rather than obfuscated user id.

* This should help debug issues with changes to how we configure
  kubectl with minikube.


jlewi:
@Jimexist Can you "/lgtm" Prow requires it.

PR #86, opened 2018-03-28
Use 2to3 and yapf to convert python code

### LINKED ISSUE #44: Verify python code is formatted with yapf
In kubeflow/tf-operator#401 we started using yapf to enforce a consistent Python style.

We should a presubmit test so that we can ensure all python code is properly formatted.

/cc @Jimexist

Jimexist:
1. make python3 syntax
1. use yapf
1. fixes #44


Jimexist:
@jlewi do you have a theory why those failed tests always produce a 404 link? I find it hard to debug (i've locally docker built the image and found nothing wrong)?

jlewi:
@Jimexist Sorry missed this ping. The 404 means the script isn't getting to the point where results are uploaded to gubernator.

The way to debug this is to trigger tests and then look for your job in the UI
https://prow.k8s.io/?repo=kubeflow%2Ftesting

And then that should allow you to get the logs.

PR #105, opened 2018-04-11
Create a new module new_test_util and other testing improvements

ankushagarwal:
Create a new module: new_test_util.py. This is intended to replace test_util.py. It includes a number of changes:
* Uses junit-xml module to create test xml files instead of manually creating them
* Show test case logs in gubernator (so that we don't have to rely on argo to show logs)
* Reduce boilerplate code required to create a new test
* Wrap every test case so that one failing test case does not prevent other tests from running
* Provide an init method which handles heavy lifting of setting up logging and common args

Also
* Refactor py_checks into test_py_checks and test_py_lint
  * These are written using new_test_util module

* Eventually the plan is to move all existing tests across Kubeflow to new_test_util

* Move from mlkube-testing to kubeflow-ci for test worker image


jlewi:
Please provide a more descriptive PR description.

ankushagarwal:
This is an experimental PR. I am experimenting with a few changes. Once I have something concrete, I will open a new PR.

PR #111, opened 2018-04-21
Create a new module: new_test_util

ankushagarwal:
Create a new module: new_test_util.py. This is intended to replace test_util.py. It includes a number of changes:

* Uses junit-xml module to create test xml files instead of manually creating them
* [Show test case logs](https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/105/kubeflow-testing-presubmit/186/) in gubernator (so that we don't have to rely on argo to show logs)
* Reduce boilerplate code required to create a new test
* Wrap every test case so that one failing test case does not prevent other tests from running
* Provide an init method which handles heavy lifting of setting up logging and common args

Also
* Refactor py_checks into test_py_checks and test_py_lint
  * These are written using new_test_util module

* Eventually the plan is to move all existing tests across Kubeflow to new_test_util

* Move from mlkube-testing to kubeflow-ci for test worker image

/cc @lluunn 
/cc @jlewi


py/kubeflow/testing/new_test_util.py
```diff
@@ -0,0 +1,155 @@
...
+class TestCase(junit_xml.TestCase):
+  """A single test case."""
+
+  def __init__(self, test_func, **kwargs):
+    self.class_name = "Kubeflow"
+    self.start_time = time.time()
+    self.test_func = test_func
+    super(TestCase, self).__init__(**kwargs)
+
+  def add_failure_info(self, message=None, output=None, failure_type=None):
```
lluunn: please add doc. what's this function for and what is each arg.

py/kubeflow/testing/new_test_util.py
```diff
@@ -0,0 +1,155 @@
...
+  """A suite of test cases."""
+
+  def __init__(self, name, artifacts_dir, logs_dir, **kwargs):
+    self.artifacts_dir = artifacts_dir
+    self.logs_dir = logs_dir
+    super(TestSuite, self).__init__(name, **kwargs)
+
+  def generate_xml(self):
+    output_file = os.path.join(self.artifacts_dir,
+                               "junit_" + self.name + ".xml")
```
lluunn: I remember we should replace "_" by "-" (or the opposite) in the name?

py/kubeflow/testing/new_test_util.py
```diff
@@ -0,0 +1,155 @@
...
+    ET.ElementTree(ET.fromstring(junit_xml.TestSuite.to_xml_string(
```
lluunn: what's this elementTree doing?

py/kubeflow/testing/new_test_util.py
ankushagarwal: I tested `_` and it seems to work.

The Gubernator example https://github.com/kubernetes/test-infra/tree/master/gubernator also uses `_`

py/kubeflow/testing/new_test_util.py
ankushagarwal: My bad. This long chained function call is not easy to understand. Added comment to explain what's happening and broke this down into 3 separate calls.

py/kubeflow/testing/new_test_util.py
ankushagarwal: Done

py/kubeflow/testing/new_test_util.py
jlewi: See #638.
https://github.com/kubeflow/kubeflow/pull/638/files#diff-2552a6a0cb5a74e3184790464405bedcR309

test-grid is the problem and we want only 1 "_" in the name.

Please fix and a comment explaining why we do this.

py/kubeflow/testing/new_test_util.py
```diff
@@ -0,0 +1,158 @@
+"""
+This module contains misc utils used for Kubeflow testing
+infrastructure. They include creating test cases/suites,
+managing logs of test cases and initializing tests
+
+This is intended to replace test_util.py
+"""
+import argparse
```
jlewi: Could we just call this package util.py?
The package is kubeflow.testing
so unless we already have a util.py file that seems fine. And this avoids the prefix "new".

We should avoid the prefix "new" because at some point we'd probably want to rename it anyway since the "new" will no longer be relevant and then we'd just have to update a bunch of call sites.

So it would be better to pick a name that we won't want to change.

jlewi:
nice change.

py/kubeflow/testing/new_test_util.py
ankushagarwal: We already have a [util.py](https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/util.py)

ankushagarwal: Renamed to `test_helper.py`

PR #122, opened 2018-05-09
Added optional param replicas

Maerville:
function wait_for_deployment may now check custom number of replicas running

Needed for https://github.com/kubeflow/kubeflow/issues/346 , as we may want to verify that all 3 replicas are Running


ankushagarwal:
Manually merging because of travis issues

PR #130, opened 2018-05-15
Update dev.kubeflow.org

jlewi:
* Update dev.kubeflow.org to use master so that we can see the latest changes.
* In particular we want to see the central UI
* recreate_app.sh shouldn't hardcode the client secret; the secret is
  created from a GCS file by redeploy_app.sh

* Update the README; the command for redeploy_app.sh was incorrect.
* Recreate the default environment because the cluster has changed.

Related to #805 Verify Central UI is working.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/130)


jlewi:
Thanks @lluunn the script to recreate the app needed to be updated to avoid deleting the components for the GitHub issue summarization example.

Issue #147, opened 2018-06-05
Test flake: Exception getting workflow status


jlewi:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/921/kubeflow-presubmit/1924/?log#log

Logs show

```
INFO|2018-06-05T19:04:02|/src/kubeflow/testing/py/kubeflow/testing/argo_client.py|21| Workflow kubeflow-presubmit-kubeflow-e2e-gke-921-4baff83-1924-8a96 in namespace kubeflow-test-infra; phase=Running
INFO|2018-06-05T19:04:02|/src/kubeflow/testing/py/kubeflow/testing/argo_client.py|21| Workflow kubeflow-presubmit-kubeflow-e2e-minikube-921-4baff83-1924-2d48 in namespace kubeflow-test-infra; phase=Running
INFO|2018-06-05T19:04:02|/src/kubeflow/testing/py/kubeflow/testing/argo_client.py|21| Workflow kubeflow-presubmit-tf-serving-image-921-4baff83-1924-9d92 in namespace kubeflow-test-infra; phase=Running
ERROR|2018-06-05T19:04:02|/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py|210| Exception occurred: 'status'
INFO|2018-06-05T19:04:03|/src/kubeflow/testing/py/kubeflow/testing/util.py|561| Writing gs://kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/921/kubeflow-presubmit/1924/finished.json
INFO|2018-06-05T19:04:03|/src/kubeflow/testing/py/kubeflow/testing/util.py|572| Uploading file /tmp/tmpRunE2eWorkflowWFHSEZlog to gs://kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/921/kubeflow-presubmit/1924/build-log.txt.
```

All the workflows completed successfully so it looks like a temporary issue getting the status

We should add some retry logic to waiting for the status so we don't fail just because we got an error.

ankushagarwal:
I have been seeing this _very_ frequently

PR #148, opened 2018-06-05
Catch and ignore exceptions in wait_for_workflows

ankushagarwal:
Retry on exceptions, except TimeoutError

Addresses https://github.com/kubeflow/testing/issues/147

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/148)

### DIFF
--- a/py/kubeflow/testing/argo_client.py
+++ b/py/kubeflow/testing/argo_client.py
@@ -21,7 +21,8 @@ def log_status(workflow):
            workflow["metadata"]["namespace"],
            workflow["status"]["phase"])
 
-@retry(stop_max_attempt_number=3, wait_fixed=2000, retry_on_exception=lambda e: not isinstance(e, util.TimeoutError))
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/argo_client.py
```diff
@@ -44,23 +44,26 @@ def wait_for_workflows(client, namespace, names,
...
+          status_callback(results)
+
+      done = True
+      for results in all_results:
+        if results["status"]["phase"] not in ["Failed", "Succeeded"]:
+          done = False
+
+      if done:
+        return all_results
+    except Exception as e: # pylint: disable=broad-except
```
jlewi: Why not use the retrying module
https://pypi.org/project/retrying/

So that we will give up after some number of retries as opposed to waiting forever.

I think we can just decorate wait_for_workflows with @retry

ankushagarwal: Done

PR #153, opened 2018-06-07
Check if status and phase is set in argo workflow crd

ankushagarwal:
Sometimes it takes a while for the argo controller to populate
the status field of an object.

Fixes https://github.com/kubeflow/testing/issues/147

/assign @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/153)

### DIFF
--- a/py/kubeflow/testing/argo_client.py
+++ b/py/kubeflow/testing/argo_client.py
@@ -66,7 +66,7 @@ def wait_for_workflows(client, namespace, names,
 
     done = True
     for results in all_results:
-      if results["status"]["phase"] not in ["Failed", "Succeeded"]:
+      if "status" not in results or "phase" not in results["status"] or results["status"]["phase"] not in ["Failed", "Succeeded"]:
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/argo_client.py
```diff
@@ -66,7 +66,10 @@ def wait_for_workflows(client, namespace, names,
 
     done = True
     for results in all_results:
-      if results["status"]["phase"] not in ["Failed", "Succeeded"]:
+      # Sometimes it takes a while for the argo controller to populate
+      # the status field of an object.
+      if "status" not in results or "phase" not in results["status"] \
```
jlewi: nit: I think its a little bit easier to read if you use get and return a suitable default
e.g
if results.get("status", {}).get("phase", "") not in ["failed", "Suceeded"]

ankushagarwal: Done

PR #155, opened 2018-06-12
Upgrade to ks 0.11 in our tests.

jlewi:
Related to kubeflow/kubeflow#727

/assign @everpeace 
/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/155)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -91,9 +91,9 @@ RUN cd /tmp && \
 # Install ksonnet
 RUN cd /tmp && \
     wget -O ks.tar.gz \
-    https://github.com/ksonnet/ksonnet/releases/download/v0.10.2/ks_0.10.2_linux_amd64.tar.gz && \
+    https://github.com/ksonnet/ksonnet/releases/download/v0.11.0/ks_0.11.0_linux_amd64.tar.gz && \
     tar -xvf ks.tar.gz && \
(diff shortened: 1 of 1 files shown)


jlewi:
New image is:
gcr.io/kubeflow-releasing/worker:v20180611-966ecb4-e3b0c4

jlewi:
/assign @pdmack 
Because he's up.

pdmack:
`ks apply will now use merge patching when it can, so Service NodePorts will not be reassigned`

This impacts us currently [here](https://github.com/kubeflow/kubeflow/pull/837#issuecomment-396062776).

jlewi:
Does that impact our tests? I've run into that as well (I think with 0.10 as well). I think that's only an issue when the service already exists; which shouldn't be the case in our tests at least.

Also I think the only NodePort service we use is for IAP which isn't used by our tests by default.

pdmack:
This was when I tried 0.11.0 out in d.k.o so you're right, perhaps just a matter of updating the app there.

jlewi:
@pdmack Do you mind approving this? I need to upgrade ks to unblock kubeflow/tf-operator#632 which is urgent because it adds CI for tf-operator v1alpha2.

pdmack:
Sure, was just poking at d.k.o to understand the ks merge changes better but I'll revisit after the commit.

/approve
/lgtm

PR #158, opened 2018-06-13
Update checkout.sh so that it merges ${PULL_BASE_REF} branch with the PR

ankushagarwal:
We then test against code which will be merged into ${PULL_BASE_REF}.
This allows fixes on ${PULL_BASE_REF} to be automatically picked up by
the PR.

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/158)

### DIFF
--- a/images/checkout.sh
+++ b/images/checkout.sh
@@ -36,7 +36,12 @@ if [ ! -z ${PULL_NUMBER} ]; then
  else
  	git checkout pr
  fi
- 
+ git merge -m "Merging origin/${PULL_BASE_REF}" origin/${PULL_BASE_REF}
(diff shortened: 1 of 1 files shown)


jlewi:
Do we really want to do this? 

I think it will be very annoying to debug PR test failures if the code in your PR doesn't match what's being test at head.

Do we have examples of where this actually caused problems?

Can we follow up with the prow team?

My understanding is that tide reruns the tests and I believe the intent is to do this after merging into master. So we can investigate how we might do this only in the tide invoked tests?

jlewi:
Should we close this issue?

PR #164, opened 2018-06-15
Add the filelock module.

jlewi:
* We run multiple tests in parallel that are modifying the ksonnet app directory
  I think this is leading to conflicts. So we want to use the filelock
  as a simple inter process communication.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/164)


ankushagarwal:
Built and pushed the image.

PR #183, opened 2018-07-23
Copy artifacts should make the file names unique by appending a suffix.

jlewi:
* See kubeflow/tf-operator#748
* A test can run multiple instances of a workflow but with different parameters.
* In this case we need to make sure the junit files and other artifacts
  copied to GCS for gubernator have unique names.

* One way to make this easier is to have copy-artifacts automatically
  append a unique suffix to each file before copying it to GCS.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/183)


kunmingg:
We can also always append a suffix liike random 6-chars string, like how k8s name its pod, and save a param.

PR #185, opened 2018-07-25
Create a new dev cluster for latest

richardsliu:
This includes the ks app files for the latest dev cluster (dev-latest.kubeflow.org). I placed them in a new directory since the other one is specific to dev.kubeflow.org, which would be the stable cluster after this change. Eventually we should also move the stable cluster ks app files here.

Also added the modified upgrade scripts (upgrade_app.sh and redeploy_app.sh), which can now take a parameter specifying "stable" or "latest".

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/185)


jlewi:
Can we also check in the Deployment Manager Configs for this Kubeflow deployment?

richardsliu:
Done.

richardsliu:
The service account json files contain private keys. Should these be pushed to github?

richardsliu:
Removed the json files with private keys.

PR #193, opened 2018-08-16
Add new configs for dev-latest.kubeflow.org via kfctl.sh

activatedgeek:
These are new configurations for the `kubeflow-latest` cluster.

/cc @richardsliu @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/193)


activatedgeek:
@richardsliu Can you verify if the current `upgrade_ks` command is enough in `deploy_latest.sh`?

deployment/README.md
```diff
@@ -1,23 +0,0 @@
-# Kubeflow deployment
```
richardsliu: Can you add the new instructions here?

dev-latest-kubeflow-org/deploy_latest.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+# Required Variables
+export KUBEFLOW_REPO=${KUBEFLOW_REPO:-}
+export PROJECT=${PROJECT:-}
```
richardsliu: We should also parameterize the kubernetes namespace. Otherwise it defaults to "kubeflow".

activatedgeek: Thanks for pointing out!

jlewi:
I have no additional comments.
This is good to go as soon as @richardsliu is satisfied.

activatedgeek:
@jlewi Do we need to have a copy of the configs here? Because these can be directly generated via `kfctl` now without a laborious process and deployed on demand.

jlewi:
Yes we need a copy of the configs.
Git should be the source of truth so there should be configs checked into source control matching what's deployed.

jlewi:
Closing this PR because I don't think it is relevant anymore.

PR #194, opened 2018-08-22
update kubernetes to 7.0.0 and websocket-client to 0.44.0 in Pipfile

ChanYiLin:
Update kubernetes 7.0.0 and websocket-client to 0.44.0 in Pipfile in image.

The reason is that kubeflow/tf-operator#790 [https://github.com/kubeflow/tf-operator/pull/790](url)
updates the kubernetes python lib to 7.0.0 since in kubernetes 7.0.0, it changed async keyword parameter in its API to async_req. 

The PR failed pre-submit-test due to our testing still use the kubernetes 4.0.0 I assume.

@jlewi  would you help to have a look

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/194)


jlewi:
I see thanks.
So this PR needs to be pushed before kubeflow/tf-operator#790. And we'll need to rebuild our testing container.

PR #210, opened 2018-09-21
Check in YAML files with current IAM policy for test and release infra projects

jlewi:
* This is the first step in modifying the current permissoins to grant
  our CI infrastructure permissions to push images to gcr.io/kubeflow-images-public

Related to:
kubeflow/testing#816 trigger docker build imagees on post-submit
kubeflow/kubeflow#1574 use prow to auto push images.

### DIFF
--- a/release-infra/README.md
+++ b/release-infra/README.md
@@ -1,0 +1,60 @@
+
+# Release Infrastructure
+
+This page describes our release infrastructure.
+
+It also contains various config files for our release infrastruce.
+
+## IAM Policies
+
+We manage IAM policies declaratively to create a versioned history of any changes.
+Changes are not automatically applied; an appropriate owner needs to apply the changes.
+
+To get the latest changes
+
+```
+gcloud projects get-iam-policy ${PROJECT} > ${PROJECT}.iam.policy.yaml
+```
+
+To change the IAM policy
+
+1. Get the current etag by running
+
+   ```
+   gcloud projects get-iam-policy ${PROJECT} > ${PROJECT}.iam.policy.yaml
+   ```
+
+1. Create and submit a PR with the desired changes
+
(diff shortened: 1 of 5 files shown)


release-infra/README.md
```diff
@@ -0,0 +1,60 @@
+
+# Release Infrastructure
+
+This page describes our release infrastructure.
+
+It also contains various config files for our release infrastruce.
+
+## IAM Policies
```
richardsliu: instrastructure

release-infra/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+To change the IAM policy
+
+1. Get the current etag by running
+
+   ```
+   gcloud projects get-iam-policy ${PROJECT} > ${PROJECT}.iam.policy.yaml
+   ```
+
+1. Create and submit a PR with the desired changes
```
richardsliu: This has the potential risk of overriding previous changes or accidentally removing someone. Would it be easier to just request access for a specific IAM role for a service account using Pantheon?

jlewi: We want Git to be the source of truth in terms of what the permissions should be. 
Otherwise its very difficult to track changes and we just accumulate cruft.

e.g. suppose someone gets added temporarily via pantheon. 6 months later how do we know whether that person still needs access?

By using Git as the source of truth, there should be a PR adding that person which explains why they were added and we can use that to infer whether access is still needed.

Ideally our presubmits would compute and show diffs with actual permissions so as part of the review process you could easily review and approve the delta.

richardsliu: My concern is that there's no way to ensure Git is the source of truth; anyone with admin access can make IAM policy changes and this file can get out of sync. I think going forward we should have some sort of an "admin playbook" for these types of issues.

Issue #207, opened 2018-09-21
Argo client should refresh credentials if OAuth token expired


richardsliu:
Currently if a workflow runs for more than an hour, the OAuth token would expire and subsequent k8s API calls would fail with a 401. A workaround is to reload the kube config when we receive an unauthorized response and retry.

Issue #204, opened 2018-09-20
[Test Flake] argo_client wait_for_workflows doesn't handle retryable errors correctly


jlewi:
Test flake:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/1579/kubeflow-presubmit/3421/?log#log

Here's the exception

```
ERROR|2018-09-20T03:41:37|/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py|257| Exception occurred: (401)
Reason: Unauthorized
HTTP response headers: HTTPHeaderDict({'Date': 'Thu, 20 Sep 2018 03:41:37 GMT', 'Audit-Id': '565e8792-b2a6-4df3-86ec-9a0e0b46ba16', 'Content-Length': '129', 'Content-Type': 'application/json', 'Www-Authenticate': 'Basic realm="kubernetes-master"'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}

Traceback (most recent call last):
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 243, in run
    status_callback=argo_client.log_status)
  File "/usr/local/lib/python2.7/dist-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/retrying.py", line 212, in call
    raise attempt.get()
  File "/usr/local/lib/python2.7/dist-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/usr/local/lib/python2.7/dist-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/src/kubeflow/testing/py/kubeflow/testing/argo_client.py", line 70, in wait_for_workflows
    GROUP, VERSION, namespace, PLURAL, n)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/apis/custom_objects_api.py", line 697, in get_namespaced_custom_object
    (data) = self.get_namespaced_custom_object_with_http_info(group, version, namespace, plural, name, **kwargs)
[...]

PR #215, opened 2018-09-21
Recreate k8s client on auth failure

richardsliu:
This is a work around for kubernetes-client/python-base#59

There is a bug in the python K8s client preventing GCP auth tokens from being refreshed when they expire.

* As a result our E2E tests fail waiting for workflows that take more than an hour. In this case, the OAuth token expires so the requests to get workspace status fail with unauthorized 

see
    #207  - Argo client should refresh credentials if OAuth token expires
   #204 - argo_client doesn't handle retries correctly.
   

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/215)

### DIFF
--- a/py/kubeflow/testing/argo_client.py
+++ b/py/kubeflow/testing/argo_client.py
@@ -29,6 +29,18 @@ def log_status(workflow):
     logging.exception('KeyError: %s', e)
 
 
+def handle_retriable_exception(exception):
+  if isinstance(exception, rest.ApiException)
+    and (exception.status == 401 or exception.status == 403):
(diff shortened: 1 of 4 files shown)


richardsliu:
/assign @jlewi 
The workaround is to (1) refresh kube config, and (2) make sure the k8s api client is recreated. For the short term fix, I'm simply recreating the client on each k8s API call.

jlewi:
Thanks so much.

/lgtm
/approve

Just a reminder please provide informative PR descriptions with links to appropriate issues. I went ahead  and updated the PR description.

PR #228, opened 2018-10-02
Allow run_e2e_workflow to select which version of ks to use

richardsliu:
- Install both ks 0.11 and 0.12 in the docker image for test-worker
- "ks" points to 0.11, "ks-12" points to 0.12
- By default, test-worker uses ks 0.11
- Run_e2e_workflow determines which version of ks to use by examining app_dir/app.yaml

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/228)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -89,13 +89,21 @@ RUN cd /tmp && \
     tar -xvf glide-v0.13.0-linux-amd64.tar.gz && \
     mv ./linux-amd64/glide /usr/local/bin/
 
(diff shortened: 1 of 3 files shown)


jlewi:
Instead of supplying the ksonnet version in prow_config.yaml; could run_e2e_workflow.py just look at 
app.yaml inside the app dir and then get the apiVersion and decide based on that which ks version to use?

apiVersion: 0.1.0

richardsliu:
Done.

PR #265, opened 2018-12-28
run_workflow_e2e.py should set the namespace on the newly created environment

jlewi:
Set the namespace of the newly created environment.
    
* Newer argo workflows will get the namespace from the environment and
  not params.


jlewi:
/cancel hold
#255 has been submitted and this has been rebased.

Issue #24, opened 2018-02-10
Get rid of one prow container per repo


jlewi:
Right now our [instructions](https://github.com/kubeflow/testing#adding-an-e2e-test-for-a-new-repository) require that every repository build a Docker image and create a custom run.sh script.

Maintaining a separate container per repository is more work than necessary. The sole purpose of this is to specify the command line arguments to run_e2e_workflow.py that control the workflow to run.

We can get rid of that by adopting a convention that the arguments will be defined in the root of the repository e.g. `/prow_config.yaml`

Then we can have a single container and entrypoint used by our prow jobs that loads the workflows and submits them.

Issue #95, opened 2018-03-31
Auto-push Kubeflow deployments corresponding to different Kubeflow versions.


jlewi:
It would be nice if we could auto-push dev.kubeflow.org or perhaps create another environment nightly-dev.kubeflow.org.

There are certain changes like UI that are difficult to test manually. It would be useful if we had an up todate environment.

I think we could easily adapt our E2E tests to do this and then just run it as a cron job.

jlewi:
It might be interesting to look at using Weave Flux
https://www.weave.works/blog/gitops-operations-by-pull-request

Weave Flux supports keeping K8s infrastructure in sync with config defined in a git repository.

Not sure if ksonnet is supported yet.

jlewi:
Here's an idea for how to make progress

1. We could use cron job to regularly run recreate_app.sh
1. Create a PR using [hub](https://github.com/github/hub) CLI for GitHub
1. Manually approve the PR

Either use argo-cd or a cron job to call redeploy_app.sh to keep the deployed instance in sync with what's checked in.

jlewi:
kubeflow/kubeflow#1100

Added a script for upgrading the ksonnet application.

jlewi:
We'd like to maintain a pool of Kubeflow deployments corresponding to different Kubeflow versions.

e.g
```
kf-vX.Y-n00
...
kf-vX.Y-n05
```

The reason for having a pool of deployments for a given major release (X.Y) is that we want to run tests against these clusters. We don't want to interrupt the tests when deploying an updated version. So by maintaining a pool and cycling through them we can create a new version for new tests while letting already running tests run to completion.

We will want to recycle the names because we want to have a fixed set of endpoints e.g

```
https://kf-v0-4-n00.endpoints.kubeflow-ci.cloud.goog/
```

The reason for having a fixed set of endpoints is that we have to manually set the redirect URIs on the OAuth credentials used for IAP. So by recycling the endpoints we don't have to manually update the OAuth credential with new redirect URIs.

#269 created a python script to deploy Kubeflow.

The next steps would be

- [ ]  Update the script in #269 to add logic to recycle the deployments
       * Check if there is an unused name `kf-vX.Y-n00,...kf-vX.Y-n05` and if there is use it
       * Otherwise delete the oldest deployment and create a new one using its name.

- [ ] Create a cron job to run the above script and redeploy using the latest commit for each release
- [ ] Create a simple flask app to provide information about all the different deployments
[...]

PR #269, opened 2018-12-28
A script to automate creation of Kubeflow clusters for testing.

jlewi:
We want to be able to automatically deploy Kubeflow from the latest commit
on branch.

This will be used for a couple purposes

  * Creating deployments that we can run our Example E2E tests against
    (e.g kubeflow/examples#92)

  * Have uptodate deployments that can be used for manually checking
    whether various features are working (e.g UIs).

The script basically invokes kfctl to deploy Kubeflow. The main addition
is to set CLIENT_ID/CLIENT_SECRET for IAP based on a file in GCS.

The deployments will be named consistently e.g.
kf-vX.y-n00,..., kf-vX.Y-n04

In a follow on PR will add logic to recycle the names; i.e. to automatically
find an unused name or delete the oldest one.

The reason we want to recycle names is because IAP requires the redirect
URIs to be manually set for the client. So by recycling the names
we can avoid having to update the credentials manually.

Related to: kubeflow/testing#24 - Use Kubeflow clusters for testing


test-infra/kf-v0-4-n00/gcp_config/cluster-kubeflow.yaml
```diff
@@ -0,0 +1,97 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
```
kunmingg: Why duplicate config files in test-infra/kf-v0-4-n00/gcp_config?
Though I believe it will be most likely stable and backward compatible.

test-infra/kf-v0-4-n00/ks_app/vendor/kubeflow/gcp/OWNERS
```diff
@@ -0,0 +1,7 @@
+approvers:
+  - abhi-g
+  - jlewi
+  - kunming
```
kunmingg: ```kunmingg```

test-infra/kf-v0-4-n00/ks_app/vendor/kubeflow/gcp/OWNERS
```diff
@@ -0,0 +1,7 @@
...
+  - llunnn
```
kunmingg: lluunn?

README.md
```diff
@@ -521,6 +521,41 @@ User or service account deploying the test infrastructure needs sufficient permi
...
+The reason we want to cycle through names is because the endpoint name for the deployment needs to be manually set in the OAuth
+credential used for IAP. By cycling through a fixed set of names we can automate redeployment without having to manually
+configure the OAuth credential.
+
+
+1. Get kfctl for the desired release
+1. Run the following command
+
+   ```
+   python -m kubeflow.testing.create_kf_instance --base_name=<kf-vX.Y|kf-vmaster>
```
kunmingg: Who will automate this call?

test-infra/kf-v0-4-n00/gcp_config/cluster-kubeflow.yaml
jlewi: kubeflow/testing#95 is tracking automating the invocation of this script.

test-infra-kf-v0-4-n00 is the kfctl app created by running kfctl. So we are just checking the configs specific to a particular kubeflow deployment.

create_kf_instance.py is just a very thin
https://github.com/kubeflow/testing/pull/269/files#diff-bfc99eeada8dcde79ce1e19a295cfb0b
wrapper around kfctl.sh

Really the only thing its doing is adding logic to
1. pick the name for the cluster 
1. Load OAuthClient ID/Secret form a GCP bucket

I agree with you we should consider refactoring the code to potentially share code. Specifically, 
we might want to reuse the logic in test_deploy_app.py to reuse SSL certificates.

I added a comment here: about resuing the logic for SSL certs.
https://github.com/kubeflow/testing/issues/95#issuecomment-450440423

I'm not sure if it makes sense to use the web app as opposed to kfctl to deploy the clusters. I think one advantage of using kfctl is that all the config files are local. With click to deploy we have to do a manual step to get access to the configs and configure them.

PR #271, opened 2018-12-30
ksonnet utilities for testing.

jlewi:
* Create a new module ks_util for common utilities related to ksonnet.

* Move the function get_ksonnet_cmd from run_e2e_workflow.py into ks_util

  * We want to reuse this function e.g. to submit TFJobs as part of E2E
    testing.

* Move setup_ks_app from
  https://github.com/kubeflow/tf-operator/blob/master/py/ks_util.py

  so it can be reused by other tests not just tf-operator.

* Update setup_ks_app to deal with different versions of ksonnet.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/271)


py/kubeflow/testing/ks_util.py
```diff
@@ -0,0 +1,55 @@
...
+import filelock
+import logging
+import os
+import re
+import subprocess
+import yaml
+
+from kubeflow.testing import util
+
+def setup_ks_app(app_dir, env, namespace, component, params, ks_cmd=None):
```
richardsliu: Will there be a separate PR to migrate other code (e.g. in tf-operator) to call this method?

PR #280, opened 2019-01-16
add `s2i` command line tool in kubeflow-ci worker image

zhenghuiwang:
[seldon serving uses source-to-image](https://github.com/SeldonIO/seldon-core/blob/master/docs/wrappers/readme.md) `s2i` to build docker images. 

This PR adds it to the CI worker image to enable e2e tests with seldon serving, e.g. xgboost-ames-housing example.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/280)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -90,6 +90,12 @@ RUN cd /tmp && \
     tar -xvf glide-v0.13.0-linux-amd64.tar.gz && \
     mv ./linux-amd64/glide /usr/local/bin/
 
+# Install s2i
+RUN cd /tmp && \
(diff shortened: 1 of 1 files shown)


zhenghuiwang:
Test: the image can be built successfully at local.
/assign @jlewi

jlewi:
/hold
/lgtm
/approve

The PR itself looks good. However, I added a hold just because there's no reason to add it to our image if we aren't going to end up using.

I think we are currently at the stage of determining whether we can run s2i inside the K8s pods or need to use GCB.

zhenghuiwang:
This PR is not longer needed because the images can be built via GCB instead inside the CI worker.
https://github.com/kubeflow/examples/pull/484

PR #286, opened 2019-01-23
Add retry block to kfctl apply all

gabrielwen:
Relates to #95

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/286)

### DIFF
--- a/py/kubeflow/testing/create_kf_instance.py
+++ b/py/kubeflow/testing/create_kf_instance.py
@@ -13,6 +13,11 @@ import yaml
 
 from google.cloud import storage
 from kubeflow.testing import util
+from retrying import retry
+
+@retry(wait_fixed=60000, stop_max_attempt_number=5)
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/create_kf_instance.py
```diff
@@ -94,7 +96,17 @@ def main(): # pylint: disable=too-many-locals,too-many-statements
   env = {}
   env.update(os.environ)
   env.update(oauth_info)
-  util.run([kfctl, "apply", "all"], cwd=app_dir, env=env)
+  # kfctl apply all might break during cronjob invocation when depending
```
jlewi: How about using the retrying module rather than implementing your own retry logic?

gabrielwen: switched to use retrying module.

PR #294, opened 2019-01-28
Update auto deploy to be runnable on kubeflow-ci.

gabrielwen:
Remaining tasks:
- Add github repos (kubeflow/testing) version snapshot.
- Dockerfile dependencies clean up.
- Deployed clusters labeling (used to identify the time deployment is made)
- Add service account role bindings garbage collection to workflow.
- Add a workflow to create git PR.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/294)


gabrielwen:
Relates to #95

py/kubeflow/testing/create_kf_instance.py
```diff
@@ -85,8 +85,12 @@ def main(): # pylint: disable=too-many-locals,too-many-statements
   # Clean up previous deployment. We are not able to run "kfctl delete all"
   # since we are not able to guarantee apps config in repository is up to date.
   util.run(["rm", "-rf", name], cwd=args.apps_dir)
-  util.run(["gcloud", "deployment-manager", "deployments", "delete", name,
-            "--project", args.project], cwd=args.apps_dir)
+  # TODO(gabrielwen):
```
jlewi: Can you file an issue for this?

Also it's probably better to use the delete_deployment script 
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/delete_deployment.sh

That script will do a more comprehensive cleanup then just deleting the deployment.

test-infra/auto-deploy/deploy-cron.yaml
```diff
@@ -13,7 +13,12 @@ spec:
         spec:
           containers:
           - name: deploy-worker
-            image: gcr.io/gabrielwen-learning/deploy-worker:latest
+            image: gcr.io/kubeflow-ci/deploy-worker:latest
```
jlewi: nit: I suggest pinning to a specific image by using a tag that won't get moved to a different image; the latest tag is typically mutable.

This way the cron job won't break just because someone builds a new image with latest.

Also, I prefer to use "command" as oppose to "args" in the pod template spec. I think this is better because it puts the full command line in one place rather than splitting it between the entrypoint defined in the dockerfile and the args soo replace this with

command:
  - /usr/local/bin/init.sh
  - /src
  - kubeflow
  ....

jlewi:
Thanks. 

nit: Would you mind updating the PR with comments explaining what changes were needed.

py/kubeflow/testing/create_kf_instance.py
gabrielwen: Done.

test-infra/auto-deploy/deploy-cron.yaml
gabrielwen: Done.

gabrielwen:
Changes include:
- Move workflows args from Dockerfile to cronjob config.  This way we don't need to rebuild the whole image for args.
- disentangle worker instantiation with github repos.  Worker init process is generic enough that we should not need to change it often - Just pull fresh copy of kubeflow and testing repo and start create_kf_instance.  This way we also don't need to rebuild worker image when we need to push out a new version - worker uses the what's on Github.
- Disable deployment deletion.  Need to figure out if it's needed and permissions to run it.

gabrielwen:
> Thanks.
> 
> nit: Would you mind updating the PR with comments explaining what changes were needed.

Added :)

PR #297, opened 2019-01-30
Github repositories snapshot and checkout.

gabrielwen:
Updates:
- Clean up dependencies.
- Take commit SHA at a timestamp and use it to set code repositories at the point.
- Write snapshot metadata to GCP Persistent Volume Claim.
- Add an argument in create_kf_instance in preparation of dynamically assignment of deployment cluster.

Relates to #95

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/297)


jlewi:
How will the git shas be used?

gabrielwen:
> How will the git shas be used?

It will be used to determine which version to be used: https://github.com/gabrielwen/testing/blob/checkout-script/test-infra/auto-deploy/checkout-snapshot.sh

jlewi:
As we discussed but I think the plan is to get rid of the init container and here's the downward API
https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

It looks like the pod labels have `job-run-name` which could be used to get a consistent id across runs

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: 2019-01-30T18:00:00Z
  generateName: label-sync-cron-1548871200-
  labels:
    controller-uid: dc982e36-24b8-11e9-a5b4-42010af001ce
    job-name: label-sync-cron-1548871200
```

Let me know if we still need this PR.

jlewi:
@gabrielwen I think as we discussed you were going to revise this. So I marked it as WIP when its ready for me to review just remove WIP from the title.

gabrielwen:
> @gabrielwen I think as we discussed you were going to revise this. So I marked it as WIP when its ready for me to review just remove WIP from the title.

I think this PR is ready :)

PR #298, opened 2019-01-31
Use snapshot files to checkout repositories.

gabrielwen:
Relates to #95 

This part uses output of #297 and checkout source codes from Github for deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/298)


kunmingg:
Do you need to do ```git config user.email```?
otherwise
/lgtm

gabrielwen:
> Do you need to do `git config user.email`?
> otherwise
> /lgtm

ran and it works.  I think you only need to mention user.email if you clone with SSH?

kunmingg:
Got it 
/approve

PR #307, opened 2019-02-07
add mercurial to test image

lluunn:
See https://github.com/kubeflow/kubeflow/issues/2364#issuecomment-461558025

Bootstrap now has a dependency that needs to be cloned by hg

/cc @kunmingg 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/307)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -26,6 +26,7 @@ RUN set -ex \
     unzip \
     gcc \
     ssh \
+    mercurial \
     python-dev \
     python-setuptools \
     python-pip \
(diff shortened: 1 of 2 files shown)


jlewi:
Where does mercurial come into play?

lluunn:
one of the dependency: bitbucket.org/ww/goautoneg
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/go.mod#L4

It's using mercurial.

PR #308, opened 2019-02-08
Add snapshot timestamp to cluster labels and use it to select deploying cluster

### LINKED ISSUE #295: Cronjob worker doesn't have permission to delete deployment
Crojob worker is not able to delete deployment with command:
gcloud deployment-manager deployments delete *

Error returned was: Deletion was aborted by user.

gabrielwen:
### Changes in this PR:
- Add snapshot timestamp to cluster labels.
- Use labels attached to select cluster to deploy.
- Add delete_deployment.sh to delete deployment before re-create it.

Fixes #295 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/308)


gabrielwen:
Fixes #295

gabrielwen:
Relates #95

py/kubeflow/testing/create_kf_instance.py
```diff
@@ -62,6 +62,14 @@ def main(): # pylint: disable=too-many-locals,too-many-statements
     "--cluster_num",
     default="", type=int, help=("Number of cluster to deploy to."))
 
+  parser.add_argument(
```
jlewi: Why are you passing this as an argument? Why don't you just have the code compute the timestamp?

gabrielwen: this timestamp is the time deployed codes taken snapshot, not the time it is deployed.

PR #317, opened 2019-02-27
Add TLS certification after creating cluster

### LINKED ISSUE #314: auto deployed KF clusters don't have http certificates because of lets encrypt quota issues.
Example

```
 curl https://kf-v0-4-n00.endpoints.kubeflow-ci.cloud.goog/
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to kf-v0-4-n00.endpoints.kubeflow-ci.cloud.goog:443 

```

zhenghuiwang:
Fixes #314 by adding self-assigned TLS certification.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/317)


images/Dockerfile
```diff
@@ -50,9 +50,11 @@ RUN cd /tmp && \
     tar -C /usr/local -xzf go.tar.gz
 
 # Install gcloud
-ENV PATH=/usr/local/go/bin:/google-cloud-sdk/bin:/workspace:${PATH} \
+ENV PATH=/root/go/bin:/usr/local/go/bin:/google-cloud-sdk/bin:/workspace:${PATH} \
     CLOUDSDK_CORE_DISABLE_PROMPTS=1
 
+RUN go get github.com/kelseyhightower/kube-rsa
```
jlewi: I think you might need to build kube-rsa. And do you want to copy the resulting binary to some location e.g. /usr/local/bin?

zhenghuiwang: The `go get` should resolve and add dependencies to the current development module and then build and install them. (from [go get doc](https://golang.org/cmd/go/#hdr-Module_aware_go_get))

I think it is easier to install package written in go this way. As a result, the binary is built and available at `/root/go/bin`. That's why I added this path to the `PATH`.

I ran shell in the new image and verified I can run `kube-rsa` directly.

jlewi: Thanks I had no idea.

PR #319, opened 2019-02-28
clean up unused bindings in iam

kunmingg:
1. fix https://github.com/kubeflow/kubeflow/issues/2570
2. delete service account only when all keys expired.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/319)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -201,19 +201,18 @@ def cleanup_service_accounts(args):
 
     keys = keys_client.list(name=a["name"]).execute()
 
-    is_expired = False
+    is_expired = True
     for k in keys["keys"]:
       valid_time = date_parser.parse(k["validAfterTime"])
       now = datetime.datetime.now(valid_time.tzinfo)
 
       age = now - valid_time
-      if age > datetime.timedelta(hours=args.max_age_hours):
-        logging.info("Deleting account: %s", a["email"])
-        iam.projects().serviceAccounts().delete(name=a["name"]).execute()
-        is_expired = True
+      if age < datetime.timedelta(hours=args.max_age_hours):
+        is_expired = False
         break
-
     if is_expired:
+      logging.info("Deleting account: %s", a["email"])
+      iam.projects().serviceAccounts().delete(name=a["name"]).execute()
       expired_emails.append(a["email"])
     else:
       unexpired_emails.append(a["email"])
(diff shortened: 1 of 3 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -201,26 +201,55 @@ def cleanup_service_accounts(args):
...
       unexpired_emails.append(a["email"])
 
   logging.info("Unmatched emails:\n%s", "\n".join(unmatched_emails))
   logging.info("Unexpired emails:\n%s", "\n".join(unexpired_emails))
   logging.info("expired emails:\n%s", "\n".join(expired_emails))
+  cleanup_service_account_bindings(unmatched_emails, unexpired_emails)
+
+
+def cleanup_service_account_bindings(unmatched_emails, unexpired_emails):
+  credentials = GoogleCredentials.get_application_default()
```
jlewi: Why not just delete all role bindings for which the service account doesn't exist?

kunmingg: Code updated.
One risk is some bindings are related to service accounts outside kubeflow-ci project. 
So filter bindings with ```is_match()```, assumes if match the related accounts are within same project.

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -222,6 +221,44 @@ def cleanup_service_accounts(args):
...
+    for a in service_accounts["accounts"]:
+      accounts.append(a["email"])
+    if not "nextPageToken" in service_accounts:
+      break
+    next_page_token = service_accounts["nextPageToken"]
+
+  resourcemanager = discovery.build('cloudresourcemanager', 'v1', credentials=credentials)
+  iamPolicy = resourcemanager.projects().getIamPolicy(resource=args.project).execute()
+  keepBindings = []
+  for binding in iamPolicy['bindings']:
```
jlewi: Does this code properly handle the case where we only delete some members for some role?

The structure of the iam policy file looks like:


```
bindings:
- members:
  - serviceAccount:593963025935@cloudbuild.gserviceaccount.com
  - serviceAccount:kubeflow-releasing@kubeflow-releasing.iam.gserviceaccount.com
  role: roles/cloudbuild.builds.builder
```
So for a given role it might be the case that we only delete some members but not others.

I think the code below will keep a binding if any member is still valid.

Why not instead just do something like

members_to_keep = []


for member in binding["members"]:
    ...
    if keep:
        members_to_keep.append(member)

binding["members"]=members_to_keep

kunmingg: Good point, updated

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -222,6 +221,46 @@ def cleanup_service_accounts(args):
...
+    members_to_keep = []
+    members_to_delete = []
```
jlewi: To make this easy to test should we break this out into a separate function which takes as input the policy yaml and returns the modified yaml.

And then add a simple test to verify its working as expected?

kunmingg: updated

jlewi:
/lgtm
/approve

Uploading a copy of the current IAM policy in case we need to restore it.

[kubeflow-ci.iam-policy.yaml.txt](https://github.com/kubeflow/testing/files/2917239/kubeflow-ci.iam-policy.yaml.txt)

PR #330, opened 2019-03-14
Cleanup firewall rules.

### LINKED ISSUE #328: Test failures; out of firewall quota
We ran out of firewall resource quota which is causing test failures.

See logs.

[kubeflow-presubmit-kfctl-2672-8dfc574-6263-3e11-2762981228.log.txt](https://github.com/kubeflow/testing/files/2964028/kubeflow-presubmit-kfctl-2672-8dfc574-6263-3e11-2762981228.log.txt)

jlewi:
* Fix #328 - Tests failing due to out of firewall quota errors.

* Add a dryrun flag to cleanup ci.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/330)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -19,6 +19,11 @@ from oauth2client.client import GoogleCredentials
 MATCHING = [re.compile("e2e-.*"), re.compile("kfctl.*"),
             re.compile("z-.*"), re.compile(".*presubmit.*")]
 
+MATCHING_FIREWALL_RULES = [re.compile("gke-kfctl-.*"),
+                           re.compile("gke-e2e-.*"),
+                           re.compile(".*presubmit.*"),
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -128,45 +135,110 @@ def cleanup_endpoints(args):
...
+
+  compute = discovery.build('compute', 'v1', credentials=credentials)
+  firewalls = compute.firewalls()
+  next_page_token = None
+
+  expired = []
+  unexpired = []
+  unmatched = []
+
+  while True:
```
gabrielwen: use retry?

jlewi: The while loop here is to handle pagination.

If we have retryable errors it would likely be on the list & delete methods. I'm not sure what errors to retry on though so my sense would be to wait until we see which errors are causing retryable failures and then deal with them.

PR #343, opened 2019-03-27
add cloud-sdk to PYTHONPATH

gabrielwen:
Relates to kubeflow/kubeflow#2190

test worker needs to use this library to check whether endpoint service is deleted.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/343)


gabrielwen:
seems like we don't need this.

PR #347, opened 2019-03-30
Fix NFS volume mounting; the NFS was recreated but still using the old IP

jlewi:
* Add a playbook for dealing with  the NFS volume being full.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/347)

### DIFF
--- a/playbook.md
+++ b/playbook.md
@@ -50,3 +50,102 @@ This is a playbook for build cops to help deal with problems with the CI infrast
      ```
      kubectl logs -l job-name=${JOBNAME}
      ```
+
+## NFS Volume Is Out Of Disk Space.
+
+1. Use [stackdriver](https://cloud.google.com/filestore/docs/monitoring-instances)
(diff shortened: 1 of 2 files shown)


playbook.md
```diff
@@ -49,4 +49,103 @@ This is a playbook for build cops to help deal with problems with the CI infrast
...
+     labels:
+     ...
+     ```
+
+     * Make sure you have the field finalizers and its an empty list
+
+   * Update the object
+
+     ```
+     kubectl apply -f /tmp/nfs-external.yaml
```
richardsliu: Should we also check in this yaml somewhere?

jlewi: It is checked in. 
https://github.com/kubeflow/testing/blob/master/test-infra/ks_app/components/nfs-external.jsonnet

That's the original spec. The spec is then modified by the controller and the finalizers are added. So we are doing a read-modify-write operation here.

PR #354, opened 2019-04-08
get_kf_testing_cluster

### LINKED ISSUE #273: E2E tests need a way to get the Kubeflow deployment they should test on
See: kubeflow/testing#95

E2E tests (such as E2E tests for our examples) need a way to request the endpoint for a Kubeflow deployment they can run.

We'd like some simply python scripts that hide the implementation details; e.g. we'd like a script that takes as input some query string that specifies constraints on the Kubeflow deployment (e.g. major version) and returns the endpoint of the cluster to use.

Under the hood this library might do something like use [Boskos](https://github.com/kubernetes/test-infra/tree/master/boskos) to request a matching cluster.

gabrielwen:
Fixes #273

Script used to inspect deployed testing clusters and returns information such as deployed time or endpoint service names.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/354)

### DIFF
--- a/py/kubeflow/testing/get_kf_testing_cluster.py
+++ b/py/kubeflow/testing/get_kf_testing_cluster.py
@@ -40,6 +40,10 @@ def list_deployments(project, name_prefix, testing_label):
 
   return cls
 
+
+def get_deployment(project, name_prefix, testing_label, desc_ordered=True):
+  pass
+
 def main(): # pylint: disable=too-many-locals,too-many-statements
   logging.basicConfig(level=logging.INFO,
                       format=('%(levelname)s|%(asctime)s'
--- a/py/kubeflow/testing/get_kf_testing_cluster.py
+++ b/py/kubeflow/testing/get_kf_testing_cluster.py
@@ -42,8 +42,12 @@ def list_deployments(project, name_prefix, testing_label, http=None, desc_ordere
       like it that HTTP requests will be made through. Should only be used in tests.
 
   Returns:
-    deployments. list. Sorted list of dictionaries with name of deployments,
-                 endpoint service name, and the timestamp the deployment is inserted.
+    deployments: list of dictionary in the format of {
+      "name": string of deployment name,
+      "endpoint": string of endpoint service name,
+      "insertTime": timestamp deployment is inserted.
+      "zone": location of deployment.
+    }
   """
   dm = None
   if http:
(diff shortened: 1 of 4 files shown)


jlewi:
Should we add some simple unittests?

Can we have a simple utility function to get the latest deployment for a particular version?

e.g. for our example tests we'd like to run on the latest cluster deployed from v0.5 or from master. Can we have a simple function to do that e.g.

get_latest(branch)

gabrielwen:
@jlewi unit tests are added and so is the new function you asked.

jlewi:
Thanks @gabrielwen 

Do the tests need other information? Should we also output the zone and cluster name?

The mnist test is currently running get-credentials
https://github.com/kubeflow/examples/blob/cc2f569ddda20ad6243fa14767451598c83be516/test/workflows/components/mnist.jsonnet#L252

Should we use Argument subparsers to add subcommands which would automatically get the latest cluster and then run run gcloud credentials? This way in our tests we could just do something like

```
get_kf_testing_cluster.py get-credentials --project=kubeflow-ci-deployment ...
```

Here's an example of using subparsers to create sub-commands that map to different functions
https://github.com/kubeflow/testing/blob/8fc539892271ae62605b9627050ef1a903160e34/py/kubeflow/testing/cleanup_ci.py#L585

gabrielwen:
> Thanks @gabrielwen
> 
> Do the tests need other information? Should we also output the zone and cluster name?
> 
> The mnist test is currently running get-credentials
> https://github.com/kubeflow/examples/blob/cc2f569ddda20ad6243fa14767451598c83be516/test/workflows/components/mnist.jsonnet#L252
> 
> Should we use Argument subparsers to add subcommands which would automatically get the latest cluster and then run run gcloud credentials? This way in our tests we could just do something like
> 
> ```
> get_kf_testing_cluster.py get-credentials --project=kubeflow-ci-deployment ...
> ```
> Here's an example of using subparsers to create sub-commands that map to different functions
> [testing/py/kubeflow/testing/cleanup_ci.py](https://github.com/kubeflow/testing/blob/8fc539892271ae62605b9627050ef1a903160e34/py/kubeflow/testing/cleanup_ci.py#L585)
> 
> Line 585 in [8fc5398](/kubeflow/testing/commit/8fc539892271ae62605b9627050ef1a903160e34)
> 
>  subparsers = parser.add_subparsers()

@jlewi 

Done, I tried to make it as flexible as possible.

PR #365, opened 2019-04-20
Add fire, gitpython and hub to test image and fix namespace packaging

jlewi:
Add fire, gitpython and hub to test image, fix namespace packaging.

* fire is a python library that makes it easy to define entrypoints
* gitpython is a python library for git; this will be used for gitops and CI/CD
* hub is a CLI for github that will be used for gitops

* Create a script to checkout repos. This is intended to be used in
  situations where we want to check out the latest code and then run some
  script.

* Fix a bug in namespace package __init__.py has a typo.

* Parameterize the makefile for the worker image so that we can build the worker image for project kubeflow-releasing using GCB.

* Add a script for configuring permissions on ssh keys.
  * This can be used in an init container to setup ssh keyes;
    see
    https://github.com/jlewi/kubeflow-dev/blob/master/notes/git_on_kubeflow.md

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/365)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/365)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -137,9 +137,9 @@ RUN cd /tmp/ && \
     pip2 install -U wheel filelock && \
     pip2 install pipenv && \
     pip2 install requests && \
-    pip2 install prometheus_client && \
+    pip2 install prometheus_client fire && \
     pipenv install --system --two && \
-    pip3 install -U wheel filelock
+    pip3 install -U wheel filelock gitpython fire
 
 RUN pip3 install pipenv==2018.10.9
 RUN cd /tmp/ && pipenv install --system --three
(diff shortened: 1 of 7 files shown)


kunmingg:
/lgtm
/hold
Who will use ```setup_ssh.sh```?

jlewi:
@kunmingg setup_ssh.sh will be used in release/testing automation for GitOps
e.g. https://github.com/kubeflow/kubeflow/pull/3080

In order to automate creation of PRs we need to be able to push to Git repos. To do that we need to setup ssh with keys.

Issue #361, opened 2019-04-19
clean up ci not clean up compute engine resources


gabrielwen:
cleanup-ci only cleans up health check resources for compute engine.  We have a much larger extent of resource leak.  Please take a look at this image:
![graph](https://storage.googleapis.com/static.ianlewis.org/prod/img/750/gcp-lb-objects2.png)

Basically we need to clean up orphaned resources plus resources associated with presubmit tests which lasted longer than a given amount of time.

This is my WIP codes for garbage collection, someone could start with it and finish it: [link](https://github.com/gabrielwen/testing/blob/garbage-collection-demo/py/kubeflow/testing/cleanup_ig.py)

Note [here](https://github.com/gabrielwen/testing/blob/fd202a547b62689fed45e4ded2bd51676abc28ab/py/kubeflow/testing/cleanup_ig.py#L31) how I use endpoint name to determine if a resource is bound to presubmit tests.  Also, this script still has a lot to improve, such as finding orphaned resources.

lluunn:
cc

lluunn:
also need to cleanup project kubeflow-ci-deployment

jlewi:
Can you provide examples of the leaked resources? 

Any idea how we are leaking instance groups? All of our instance groups should be associated with GKE clusters. So if we're leaking instance groups that suggests deleting GKE cluster isn't deleting the associated instance group.

jlewi:
Attached are list of the current instance groups and clusters in project kubeflow-ci-deployment

A lot of the instance groups have size zero.

The instance groups don't look to be named after our E2E test clusters; e.g.

```
k8s-ig--f16b03f306d1329d
```

Some of them appear to be in use by forwarding rules. 

One hypothesis is that these are node pools provisioned by auto-provisioning and that might explain why they aren't garbage collected.

[clusters.list.txt](https://github.com/kubeflow/testing/files/3098082/clusters.list.txt)
[instance_groups.list.txt](https://github.com/kubeflow/testing/files/3098083/instance_groups.list.txt)

jlewi:
Here are the stackdriver logs for one instance group. It was created by the GKE service account.


[...]

PR #382, opened 2019-05-04
GC backend services in kubeflow-ci-deployment

zhenghuiwang:
#361 

Clean up backend services older than 24 hours in kubeflow-ci-deployment. The health checks related to the backend service will be cleaned up by existing `cleanup_health_checks` function.

/cc @jlewi @gabrielwen

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/382)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -253,6 +253,44 @@ def cleanup_firewall_rules(args):
   logging.info("Unexpired firewall rules:\n%s", "\n".join(unexpired))
   logging.info("expired firewall rules:\n%s", "\n".join(expired))
 
+def cleanup_backend_services(args):
+  # We only GC backend services for kubeflow-ci-deployment.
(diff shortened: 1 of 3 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -253,6 +253,44 @@ def cleanup_firewall_rules(args):
   logging.info("Unexpired firewall rules:\n%s", "\n".join(unexpired))
   logging.info("expired firewall rules:\n%s", "\n".join(expired))
 
+def cleanup_backend_services(args):
```
gabrielwen: shouldn't you check if a given backend service is used by some health checks?  If it is, I think deletion will be rejected, no?

zhenghuiwang: A backendservice can be deleted even if it is used by health checks. that's why we have cleanup_health_checks for orphan health checks.

zhenghuiwang:
the presumit test actually succeeds.

/retest

PR #378, opened 2019-04-30
Use the injected base branch env var

derekhh:
**Background**

This fixes #375.

**Changes**

Used `PULL_BASE_REF` to get the base branch name and get a common ancestor for the PR and the base branch.

**Tests**

In Progress.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/378)

PR #385, opened 2019-05-07
Fetch remote branches in E2E test

derekhh:
**Background**

In https://github.com/kubeflow/kubeflow/pull/3216, @gabrielwen is observing error messages like ```fatal: Not a valid object name v0.5-branch``` in the presubmit tests.

This is because I was doing things like `git merge-base HEAD v0.5-branch` in https://github.com/kubeflow/testing/pull/378, e.g.

```
~/src/kubeflow/kubeflow master
❯ git merge-base HEAD v0.5-branch
fatal: Not a valid object name v0.5-branch
```

Instead we should do:

```
~/src/kubeflow/kubeflow v0.5-branch
❯ git merge-base HEAD remotes/origin/v0.4-branch
d48e22b2f7d7b80e14c5a07bf171f358b9f91131
```

**Changes**

* Added `remotes/origin` before the branch name in `git merge-base`

**Tests**

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/385)

### DIFF
--- a/py/kubeflow/testing/run_e2e_workflow.py
+++ b/py/kubeflow/testing/run_e2e_workflow.py
@@ -125,8 +125,9 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
   if job_type == "presubmit":
     # We need to get a common ancestor for the PR and the base branch
     cloned_repo_dir = os.path.join(args.repos_dir, repo_owner, repo_name)
-    _ = util.run(["git", "checkout", base_branch_name], cwd=cloned_repo_dir)
-    common_ancestor = util.run(["git", "merge-base", "HEAD", base_branch_name],
+    _ = util.run(["git", "fetch", "origin", base_branch_name], cwd=cloned_repo_dir)
+    common_ancestor = util.run(["git", "merge-base", "HEAD",
+                                "remotes/origin/{}".format(base_branch_name)],
                                cwd=cloned_repo_dir)
     diff_command = ["git", "diff", "--name-only", common_ancestor]
   elif job_type == "postsubmit":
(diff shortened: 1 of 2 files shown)


py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -124,9 +124,10 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
   diff_command = []
   if job_type == "presubmit":
     # We need to get a common ancestor for the PR and the base branch
-    common_ancestor = util.run(
-      ["git", "merge-base", "HEAD", base_branch_name],
-      cwd=os.path.join(args.repos_dir, repo_owner, repo_name))
+    cloned_repo_dir = os.path.join(args.repos_dir, repo_owner, repo_name)
+    _ = util.run(["git", "checkout", base_branch_name], cwd=cloned_repo_dir)
```
jlewi: Do you want to checkout base_branch_name or just fetch it?

If you check it out then I think the current version of the checked out code would be set to v0.5-branch which isn't what we want since we want the current branch to be the branch corresponding to the PR.

derekhh: Thanks for pointing this out! Yeah I should just do `git merge-base remotes/origin/blah` instead of a `git checkout` as this changes the HEAD.

Now I'm doing a `git fetch origin blah` to prevent the case where `remotes/origin/blah` may not exist locally (I can't think of a case though but just doing that for safety) and then `git merge-base remotes/origin/blah`.

PTAL?

derekhh:
@jlewi - PTAL? Thanks! :)

jlewi:
Thanks
/lgtm
/approve

PR #390, opened 2019-05-09
Cleanup all ingress resources

### LINKED ISSUE #361: clean up ci not clean up compute engine resources

zhenghuiwang:
Fixes #361 

Clean up all ingress resources after certain age:
-  cleanup_forwarding_rules,
-  cleanup_target_http_proxies,
-  cleanup_url_maps,
-  cleanup_instance_groups,

Tested it locally.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/390)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -611,7 +787,11 @@ def cleanup_all(args):
          cleanup_service_account_bindings,
          cleanup_workflows,
          cleanup_disks,
+         cleanup_forwarding_rules,
+         cleanup_target_http_proxies,
+         cleanup_url_maps,
          cleanup_backend_services,
+         cleanup_instance_groups,
```
gabrielwen: I think this one needs to be after `cleanup_health_checks`.

zhenghuiwang: Done.

zhenghuiwang:
> have you tried and see if there is any error because of dependencies?

I don't see much error about. I wouldn't worry about the order of these clean up operations. It is a cron job and the resource will be GCed once the dependencies are GCed in last round.

gabrielwen:
> > have you tried and see if there is any error because of dependencies?
> 
> I don't see much error about. I wouldn't worry about the order of these clean up operations. It is a cron job and the resource will be GCed once the dependencies are GCed in last round.

theoretically yes, but it would take O(n) rounds to GC everything, given n = level of dependencies, right?  we are already witnessing increasing speed of resource consumption, should we make some effort to reduce the time of GC?

zhenghuiwang:
Added a comment on the order of these operations, which now clean all ingress resources in one run. PTAL

PR #431, opened 2019-08-02
[WIP] Support Python functions in workflows

scottilee:
@jlewi I hope this is somewhat right... please help me fix and iterate on this 😅

### DIFF
--- a/py/kubeflow/testing/run_e2e_workflow.py
+++ b/py/kubeflow/testing/run_e2e_workflow.py
@@ -3,7 +3,7 @@
 This script submits Argo workflows to run the E2E tests and waits for
 them to finish. It is intended to be invoked by prow jobs.
 
-It requires the workflow to be expressed as a ksonnet app.
+It requires the workflow to be expressed as a ksonnet app or a Python function.
 
 The script can take a config file via --config_file.
 The --config_file is expected to be a YAML file as follows:
@@ -17,9 +17,9 @@ workflows:
     include_dirs:
       tensorflow/*
 
-  - name: lint
-    app_dir: kubeflow/kubeflow/testing/workflows
-    component: workflows
+  - py_func: echo
+     kw_args:
+         arg1: hello
 
 app_dir is expected to be in the form of
 {REPO_OWNER}/{REPO_NAME}/path/to/ksonnet/app
@@ -32,6 +32,10 @@ that should run this workflow.
 
 include_dirs (optional) is an array of strings that specify which directories, if modified,
 should run this workflow.
+
+py_func is the Python method to invoke Argo workflows
+
+kw_args is an array of arguments passed to the Python method
 
 The script expects that the directories
 {repos_dir}/{app_dir} exists. Where repos_dir is provided
@@ -62,7 +66,7 @@ def get_namespace(args):
     return "kubeflow-releasing"
   return "kubeflow-test-infra"
 
-class WorkflowComponent(object):
+class WorkflowKSComponent(object):
   """Datastructure to represent a ksonnet component to submit a workflow."""
 
   def __init__(self, name, app_dir, component, job_types, include_dirs, params):
@@ -72,6 +76,13 @@ class WorkflowComponent(object):
     self.job_types = job_types
     self.include_dirs = include_dirs
     self.params = params
+
+class WorkflowPyComponent(object):
+  """Datastructure to represent a Python component to submit a workflow."""
+
+  def __init__(self, py_func, kw_args):
+    self.py_func = py_func
+    self.args = [v for k, v in kw_args.items()]
 
 def _get_src_dir():
   return os.path.abspath(os.path.join(__file__, "..",))
@@ -89,9 +100,13 @@ def parse_config_file(config_file, root_dir):
 
   components = []
   for i in results["workflows"]:
-    components.append(WorkflowComponent(
-      i["name"], os.path.join(root_dir, i["app_dir"]), i["component"], i.get("job_types", []),
-      i.get("include_dirs", []), i.get("params", {})))
+    if i.get("app_dir"):
+      components.append(WorkflowKSComponent(
+        i["name"], os.path.join(root_dir, i["app_dir"]), i["component"],
+        i.get("job_types", []), i.get("include_dirs", []), i.get("params", {})))
+    if i.get("py_func"):
+      components.append(WorkflowPyComponent(
+        i["py_func"], i.get("kw_args", {})))
   return components
 
 def generate_env_from_head(args):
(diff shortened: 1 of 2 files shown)


jlewi:
Is this ready for review?
If its ready for review you should remove WIP from the title and make sure you "/assign" it to the reviewers?

scottilee:
/assign @jlewi 

I made some changes based on your feedback. Why does the `create_workflow` method need to return a dictionary? Is it okay for it to return yaml like I did and just create the workflow object?

Also, I'm not sure how to handle the args... let me know the how to handle them.

jlewi:
create_workflow should return a Python object so that its easier for other python code to manipulate. The alternative would be a string containing the YAML definition of the workflow.

Your code is returning a pythong

```
  yaml_result = yaml.safe_load(argo_hello_world.text)
  return yaml_result
```

jlewi:
@scottilee It doesn't look like your test is actually running the workflows. I suspect there is an error in run_e2e_workflow.py and your script is aborting.

You want to look at the logs for your prow jobs; some directions here
https://github.com/kubeflow/testing#prow

Here's a list of the prow jobs for kubeflow/testing
https://prow.k8s.io/?repo=kubeflow%2Ftesting

So after you trigger a presubmit run
You should be able to look at 
https://prow.k8s.io/?repo=kubeflow%2Ftesting

And see your prow job. From there you can get the pod logs for the prow job; which should show you stdout/stderr for run_e2e_workflow.py

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -89,9 +101,13 @@ def parse_config_file(config_file, root_dir):
...
+    if i.get("app_dir"):
+      components.append(WorkflowKSComponent(
+        i["name"], os.path.join(root_dir, i["app_dir"]), i["component"],
+        i.get("job_types", []), i.get("include_dirs", []), i.get("params", {})))
+    if i.get("py_func"):
+      components.append(WorkflowPyComponent(
+        i["py_func"], i.get("kw_args", {})))
   return components
 
 def generate_env_from_head(args):
```
kkasravi: On line 200
```
    workflow_name = os.getenv("JOB_NAME") + "-" + w.name
    ks_cmd = ks_util.get_ksonnet_cmd(w.app_dir)
```
should move to under line 259
```
    if w.app_dir:
```

jlewi:
@scottilee When do you think you might have another version of this PR ready for review.

It would be great to get this working. It will unblock a lot of work.

py/kubeflow/testing/run_e2e_workflow.py
scottilee: Can you explain this? `workflow_name` is being used between line 201 and line 252.

kkasravi: either py_func needs a name as well or we use py_func in place of w.name

scottilee:
@jlewi just pushed another update.

> So after you trigger a presubmit run
> You should be able to look at
> https://prow.k8s.io/?repo=kubeflow%2Ftesting
> 
> And see your prow job. From there you can get the pod logs for the prow job; which should show you stdout/stderr for run_e2e_workflow.py

I'm not seeing mine so I'm assuming if I wait more than a day it will be gone?


> create_workflow should return a Python object so that its easier for other python code to manipulate. The alternative would be a string containing the YAML definition of the workflow.
> 
> Your code is returning a pythong
> 
> ```
>   yaml_result = yaml.safe_load(argo_hello_world.text)
>   return yaml_result
> ```

I forget why I said earlier that `create_workflow` is returning YAML but it is actually returning a Python object (a dictionary). This is what you want to happen, correct?

py/kubeflow/testing/run_e2e_workflow.py
scottilee: Ah, I think I'm understanding what you're saying, ha.

scottilee: Fixed.

scottilee: Sigh, I thought I fixed it. I don't get why it's still throwing an error since it shouldn't include `WorkflowPyComponent` given that the workflow doesn't have an `app_dir` property...

```
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 196, in run
    if w.app_dir:
AttributeError: 'WorkflowPyComponent' object has no attribute 'app_dir'
```

jlewi: It might be worth updating the unittest so we verify part of the code is working just by running a unittest
https://github.com/kubeflow/testing/blob/master/py/kubeflow/tests/run_e2e_workflow_test.py

Refactoring the code may help.

In particular; IIUC the part of code that is causing problems is converting a prow_config.yaml file into a set of workflows and then verifying in the case of PyFunc we can extract a dictionary containing the expected Argo workflow.

Can we validate that using a unittest?

jlewi: Expanding on that a bit. I might suggest narrowing the scope of this first PR to just implement the following function

```
def build_workflows(prow_config):
   """
       Returns: 
           workflows: List of dictionarys containing Argo workflows
    """
    ....
```

* This function would take a dictionary containing the contents of the prow_config.yaml file and return a list of dictionaries containing argo workflows
* This function should only handle py_func workflows; entries defining ksonnet based workflows should just be skipped

In follow on PRs we can integrate this into the overall workflow. I think we can do that by refactoring the code to something like the following

```
def run(...)
  ... various setup ...
   prow_config = load prow_config
   ks_workflows = submit_ksonnet_workflows(prow_config)
   py_workflows = build_workflows(prow_config)
   submit_argo_workflows(py_workflows)

   all_workflows = ks_workflows + py_workflows
   wait_for_all_workflows(all_workflows)
```\
* So basically we would refactor the ksonnet specific bits of run into submit_ksonnet_workflows

**Comment by scottilee** [comment]
Thanks to @kkasravi the test has progressed but now it's failing at line 313 saying that `create_started_file` is invalid syntax? Not sure why this is happening.

Screen Shot 2019-08-16 at 12 33 14 AM

**Comment by kkasravi** [comment]
You should run with python3

**Comment by jlewi** [comment]
I believe that's set here.
https://github.com/kubeflow/testing/blob/8cbd626f678d2c830edc38d5990bc2e1c6e1eb70/images/run_workflows.sh#L12

If we need to change that to python3 we will need to build a new image and verify that its not breaking anything.

Could we stick with python2 syntax or make the code python2 compatible so we don't have to upgrade for this PR?

**Comment by jlewi** [comment]
@scottilee Any thought about my comment above about refactoring the code so its easy to write a unittest to ensure the code is returning the expected workflow?

**Comment by scottilee** [comment]
@kkasravi I updated your function to include the args and for it to use the Python2 version of importlib (imp). Still getting the same error from the screenshot so I don't think it has to do with the Python version. Any other ideas?

```
  File "/usr/lib/python2.7/runpy.py", line 163, in _run_module_as_main
    mod_name, _Error)
  File "/usr/lib/python2.7/runpy.py", line 119, in _get_module_details
    code = loader.get_code(mod_name)
  File "/usr/lib/python2.7/pkgutil.py", line 283, in get_code
    self.code = compile(source, self.filename, 'exec')
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 316
    create_started_file(args.bucket, ui_urls)
                      ^
SyntaxError: invalid syntax
```

@jlewi I'll look at the unittest later today, I'll probably need some help though.

**Comment by kkasravi** [comment]
@jlewi, i’ve been implementing Tektoncd pipelines using python3 as well. I built a new image of testworker and have been running in a local k8 cluster or using docker

**Comment by jlewi** [comment]
@kkasravi Should we try to rollout these changes 1 by 1 e.g

1. Add support for py_func
1. Add support for tektoncd
1. migrate from python2 to python3

**Comment by kkasravi** [comment]
Ok, though I would move 3 to 1

**Comment by jlewi** [comment]
if we migrate to python3 that potentially affects all of our tests. Right now we are blocked on not having a good way to write E2E tests that don't use ksonnet. I'd really like to avoid doing any migrations changes until we have an alternative solution for ksonnet for writing E2E tests.

Why do we need to migrate to Python3 to support python functions? What is the code that requires python3?  Can we write that in a way that is forwards compatible with python 3; e.g. guard it with an if else that says what to do for py2 and what to do for py3?

**Comment by kkasravi** [comment]
i believe @scottilee is doing everything in python2, originally i had used a module loader that we thought was python3 specific. He updated me today that this wasn't the case. So we can move the priority of python2->python3 back to (3).

**Comment by scottilee** [comment]
@jlewi below is a screenshot from the most recent failure from Prow. `ci/kf_unittests.py` is a new file I created per your earlier instructions on a comment above. Is there something I need to do to make sure the presubmit job picks it up? The weird thing is that when I test this locally I don't get the same import error. Most likely the paths on my system are different from what the CI system is using but not sure how to reconcile it.

Screen Shot 2019-08-19 at 2 47 03 PM

Also I looked at the `run_e2e_workflow_test.py` file. Honestly I haven't done much test writing before so a lot of it is gibberish to me... not confident I can update it, sorry.

Also feel free to jump in more, I definitely don't want to keep holding you and Kam back on this 🙁

**Comment by jlewi** [comment]
The convention is to add the PY directory to the python path
https://github.com/kubeflow/testing/tree/master/py/

So would generally import this as

```
import kubeflow.testing.ci
```

**Comment by jlewi** [comment]
@scottilee Didn't the file kf_unittests.py already exist? It looks like that function is still trying to load an argo wokflow from a yaml file and parse it.

The suggestion I made above was to define the function

```
def build_workflows(prow_config):
   """
       Returns: 
           workflows: List of dictionarys containing Argo workflows
    """
    ....
```

Lets implement that function and a unittest. Then you should be able to run the unittest in your favorite debugger and verify that you get the workflows are expected. 
Once that function is working and the test is passing we can merge that function and then in a follow on PR rewrite run_e2e_workflow.py to use it.

scottilee:
@jlewi presubmit tests finally passed!

jlewi:
Woo Hoo!

jlewi:
The workflow isn't showing up in gubernator
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/431/kubeflow-testing-presubmit/1164044441279795200/

Will take a look at the code; might be worth punting on to a follow on PR.

jlewi:
Thanks @scottilee please take a look at my comments in revieable I highlighted a couple fixes that should cause the workflows to show up in gubernator.

scottilee:
@jlewi I apologize for the delay on this.

1. Could you clarify what you mean by your comment "This"?
2. Seeing as how the py_func doesn't have a name property, what should I append to the `workflow_array` list?

jlewi:
@scottilee I think you need to look at my comments in reviewable
https://reviewable.io/reviews/kubeflow/testing/431

GitHub is only showing you a portion of the comment which is why I think its just showing you "This".

Your py_func returns an disctionary defining an Argo workflow. You will wan to set metadata.Name in that workflow by using logic similar to what we currently have to generate the workflow_name
https://github.com/kubeflow/testing/blob/54db950f9888b3da0295e6b230ce4353476e3edd/py/kubeflow/testing/run_e2e_workflow.py#L216

You'll need to set metadata.Name in the Argo workflow so that the workflow will be created with the desired name. You'll then want to append it to the list argo workflows so that the code will wait for that workflow to be created.

jlewi:
@scottilee Do you think you will have time to tackle this early this week (I know this isn't part of your full time job)? 

We could submit this as is and then you or someone else could make the additional fixes in a follow on PR.

If you think you'll have time to take another stab at this PR today or tomorrow then I think we should wait.

PR #432, opened 2019-08-05
upgrade tornado to fix fairing e2e test failing

jinchihe:
Upgrade tornado to fix integration test failure https://github.com/kubeflow/fairing/issues/336 in Kubeflow Fairing.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/432)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -148,7 +148,7 @@ RUN cd /tmp/ && \
     pip2 install requests && \
     pip2 install prometheus_client fire && \
     pipenv install --system --two && \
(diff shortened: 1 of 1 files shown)


jlewi:
Is it possible to update the E2E tests in this repo to use the new image so that we know its working.
maybe if we set it here
https://github.com/kubeflow/testing/blob/52615d4bb26558db7d6f5abcac20aebac465a99c/workflows/components/workflows.libsonnet#L37

jinchihe:
/close

The problem has been fixed in the Fairing, thanks.

PR #434, opened 2019-08-10
[WIP] run_e2e_workflow.py support TektonCD pipeline

### LINKED ISSUE #425: run_e2e_workflow.py support TektonCD pipeline
We should modify run_e2e_workflow.py so that we can also submit and run tektoncd/pipeline's and not just argo workflows.

We probably want to fork our argo_client to create a tekton client
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/argo_client.py

blocked by kubeflow/testing#423 run_e2e_workflow.py should support calling py functions to generate the workflow.

/cc @scottilee

kkasravi:
fixes #425

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/434)


jinchihe:
@kkasravi @scottilee if we need to consider compatibling with current testing for smooth transition? if we removed all old ksonnet logic, the kubeflow repos PR will cannot pass testing until update the CI/CD testing such as `prow_config.yaml`. Thanks!

kkasravi:
@jinchihe we'll keep the ksonnet logic until 'app_dir' doesn't exist in any prow_config.yaml

PR #441, opened 2019-08-15
Fix auto-deployment from master

jlewi:
* Auto deployment from master should use --config flag to initialize KfDef
  * I think we are pulling in outdated configs which is why master deployment
    isn't working

* Clean up some of the old code for deploying with kfctl.sh since we no longer
  use that.

* Related to kubeflow/kubeflow#3905

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/441)

### DIFF
--- a/test-infra/auto-deploy/deploy-master.yaml
+++ b/test-infra/auto-deploy/deploy-master.yaml
@@ -30,6 +30,7 @@ spec:
         - --max_num_cluster=5
         - --zone=us-east1-b
         - --github_token_file=/secret/github-token/github_token
+        - --kfctl_config=https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_gcp_iap.yaml
         volumeMounts:
         - name: gcp-credentials
           mountPath: /secret/gcp-credentials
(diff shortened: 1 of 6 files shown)


jlewi:
@lluunn I think this is ready for review.
The scripts appear to be deploying from master. 
The endpoints still aren't accessible but I think that's an issue with master since the postsubmits are also failing
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-postsubmit

/assign @lluunn

test-infra/auto-deploy/deploy-master.yaml
```diff
@@ -22,14 +22,16 @@ spec:
           value: /secret/gcp-credentials/key.json
         command:
         - /usr/local/bin/auto_deploy.sh
-        - --repos=kubeflow/kubeflow;kubeflow/testing
+        #- --repos=kubeflow/kubeflow;kubeflow/testing
+        - --repos=kubeflow/kubeflow;jlewi/testing@fix_nfs2
```
zhenghuiwang: Do we need to change this back to `kubeflow/testing` after merging this pr?

jlewi: Done.

jlewi:
@lluunn Could you re-LGTM and approve please.

PR #442, opened 2019-08-16
Updating labels to use kind prefix

### LINKED ISSUE #440: Standardize/cleanup different bug kind labels
The following labels are currently used to indicate what kind an issue is

* community/discussion
* community/question
* kind/bug
* improvement/enhancement
* addition/feature
* kind/process

Not  using the kind prefix has a bunch of drawbacks

* Prow allows the "/kind" command to be used to set the kind.
  * We use this in our issue templates to automatically set the kind
  * If we don't use kind as a prefix people can't set it via the kind bot
* It makes our [issue triage process](https://github.com/kubeflow/community/blob/master/proposals/issue_triage.md) more complicated because we'd like to set different requirements based on the kind
   * e.g. A discussion probably doesn't need a priority or be assigned to a project

I propose we use the following kinds

* kind/bug
* kind/question
* kind/feature
* kind/discussion
* kind/process

I think we want to make the following changes

* Update [kubeflow_label.yml](https://github.com/kubeflow/testing/blob/master/label_sync/kubeflow_label.yml) 

   * Map the old times to the new types

* Update the running instance of label sync and make sure it succesfully applies across all repos
* Update the issue label bot configs ([example](https://github.com/kubeflow/kubeflow/blob/master/.github/issue_label_bot.yaml)) in all repos so that it uses the new kinds

rbrishabh:
Fixes #440 

Hey @jlewi !

Let me know if this needs any changes!

Thanks! =)

### DIFF
--- a/label_sync/kubeflow_label.yml
+++ b/label_sync/kubeflow_label.yml
@@ -46,13 +46,22 @@ default:
   - color: d2b48c
     name: area/usage
   - color: ecfc15
-    name: community/discussion
+    name: kind/discussion
+    previously:
+    - name: community/discussion
   - color: ecfc15
     name: community/maintenance
   - color: 2515fc
-    name: community/question
+    name: kind/question
     previously:
+    - name: community/question
     - name: question
+  - color: 2515fc
+    name: kind/feature
+    previously:
+    - name: addition/feature
+  - color: 2515fc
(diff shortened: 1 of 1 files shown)


rbrishabh:
@googlebot I signed it!

PR #448, opened 2019-08-28
fix presubmit

lluunn:
for https://github.com/kubeflow/kubeflow/issues/4032

We should only add workflow to the list when it's being triggered.

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/448)


lluunn:
not sure what's this if else doing:
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py#L230

PR #451, opened 2019-08-29
Support Python functions in workflows

### LINKED ISSUE #423: run_e2e_workflow.py should support using python functions to generate the workflow
Currently we use ksonnet to parameterize our Argo workflows in our E2E tests.
Since ksonnet is being deprecated we need a different way to parameterize our E2E tests.
I think a flexible way to do this is to let people define a class in python with methods that return one or more Argo workflows.

```
class MyTestWorkflows(object):
   def workflow1(self, param1=1)
        ...
        return workflow
....
```

Our prow_config.yaml files could then look like the following

```
workflows
- py_func: package.module.class.method_name
   kw_args:
       arg1: value
```

Then run_e2e_workflow.py can 
   1. Load prow_config.yaml
   1. Invoke py_func with the specified arguments
   1. submit and wait for the supplied Argo workflow

just invoke `MyTestWorkflows` to get the workflow.

When we make this change in the same PR we should replace the existing ksonnet test in kubeflow/testing. This way the presubmit will verify that the new code is working.

Related to: kubeflow/kubeflow#3035 migrate off ksonnet

@scottilee Any interest in picking this up?

scottilee:
As ksonnet is being deprecated, this PR allows for Python functions to be used to parameterize Argo workflows in E2E tests.

Fixes #423 

Related to https://github.com/kubeflow/testing/pull/431 which was rolled back. This should be fixed as with the following section which is moved to the correct place:

```
    # Skip this workflow if it is scoped to a different job type.
    if w.job_types and not job_type in w.job_types:
      logging.info("Skipping workflow %s because job type %s is not one of "
                   "%s.", w.name, job_type, w.job_types)
      continue
```

/assign @jlewi 
/cc @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/451)

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,6 +680,8 @@ Some examples to look at
 
 Follow these steps to add a new test to a repository.
 
+#### ksonnet
+
 1. Create a ksonnet App in that repository and define an Argo workflow if one doesn't already exist
    * We use ksonnet apps defined in each repository to define the Argo workflows corresponding to E2E tests
    * If a ksonnet app already exists you can just define a new component in that app
@@ -758,6 +760,32 @@ Follow these steps to add a new test to a repository.
 
      * **params**: A dictionary of parameters to set on the ksonnet component e.g. by running `ks param set ${COMPONENT} ${PARAM_NAME} ${PARAM_VALUE}`
 
+#### Python function
+
+* **Note**: ksonnet is being deprecated, please start using Python functions to parameterize Argo workflows in E2E tests.
+
+1. Create a Python function in that repository and return an Argo workflow if one doesn't already exist
+   * We use Python functions defined in each repository to define the Argo workflows corresponding to E2E tests
+
+   * You can look at `prow_config.yaml` (see below) to see which Python functions are already defined in a repository.
+
+1. Modify the `prow_config.yaml` at the root of the repo to trigger your new test.
+
+   * If `prow_config.yaml` doesn't exist (e.g. the repository is new) copy one from an existing repository ([example](https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml)).
+
+   * `prow_config.yaml` contains an array of workflows where each workflow defines an E2E test to run; example
+
+       ```
+       workflows:
+        - name: workflow-test
+          py_func: my_test_package.my_test_module.my_test_workflow
+          kw_args:
+              arg1: argument
+       ```
+
+       * **py_func**: Is the Python method to invoke Argo workflows
+       * **kw_args**: This is an array of arguments passed to the Python method
+       * **name**: This is the base name to use for the submitted Argo workflow.
 
 ### Prow Variables
 
--- a/README.md
+++ b/README.md
@@ -783,7 +783,7 @@ Follow these steps to add a new test to a repository.
               arg1: argument
        ```
 
-       * **py_func**: Is the Python method to invoke Argo workflows
+       * **py_func**: Is the Python method to create a python object representing the Argo workflow resource
        * **kwargs**: This is an array of arguments passed to the Python method
        * **name**: This is the base name to use for the submitted Argo workflow.
 
--- a/README.md
+++ b/README.md
@@ -779,12 +779,12 @@ Follow these steps to add a new test to a repository.
        workflows:
         - name: workflow-test
           py_func: my_test_package.my_test_module.my_test_workflow
-          kw_args:
+          kwargs:
               arg1: argument
        ```
 
        * **py_func**: Is the Python method to invoke Argo workflows
-       * **kw_args**: This is an array of arguments passed to the Python method
+       * **kwargs**: This is an array of arguments passed to the Python method
        * **name**: This is the base name to use for the submitted Argo workflow.
 
 ### Prow Variables
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -4,3 +4,10 @@ workflows:
   - app_dir: kubeflow/testing/workflows
     component: workflows
     name: unittests
+
+  - py_func: kubeflow.testing.ci.kf_unittests.create_workflow
+    name: pyfunctest
+    # can optionally take keyword arguments
+    # kw_args:
+    #  a: 1
+    #  b: 2
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -5,6 +5,18 @@ workflows:
     component: workflows
     name: unittests
 
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: click_deploy_test
+    name: skipmeplease
+    job_types:
+    - periodic
+    include_dirs:
+    - bootstrap/*
+    - kubeflow/*
+    - testing/*
+    params:
+      workflowName: skipmeplease
+
   - py_func: kubeflow.testing.ci.kf_unittests.create_workflow
     name: pyfunctest
     # can optionally take keyword arguments
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -5,18 +5,19 @@ workflows:
     component: workflows
     name: unittests
 
-  - app_dir: kubeflow/kubeflow/testing/workflows
+  # Verifies the workflow is skipped because the test would fail since it
+  # doesn't exist
+  - app_dir: kubeflow/kubeflow/testing/lmnop
     component: click_deploy_test
     name: skipmeplease
     job_types:
-    - periodic
+    - presubmit
     include_dirs:
-    - bootstrap/*
-    - kubeflow/*
-    - testing/*
+    - xyz/*
     params:
       workflowName: skipmeplease
 
+  # Smoke test to verify that the workflow is actually triggered
   - py_func: kubeflow.testing.ci.kf_unittests.create_workflow
     name: pyfunctest
     kwargs:
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -19,7 +19,5 @@ workflows:
 
   - py_func: kubeflow.testing.ci.kf_unittests.create_workflow
     name: pyfunctest
-    # can optionally take keyword arguments
-    # kw_args:
-    #  a: 1
-    #  b: 2
+    kwargs:
+     url: https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml
--- a/py/kubeflow/testing/ci/kf_unittests.py
+++ b/py/kubeflow/testing/ci/kf_unittests.py
@@ -1,0 +1,15 @@
+"""Test Argo workflow"""
+
+import requests
+import yaml
+
+def create_workflow():
+  """ Loads Argo example YAML and returns dictionary object """
+  # TODO: define workflow to run unittests
+  argo_hello_world = requests.get("https://raw.githubusercontent.com/argoproj/"
+  "argo/master/examples/hello-world.yaml")
+  yaml_result = yaml.safe_load(argo_hello_world.text)
+  return yaml_result
+
+if __name__ == "__main__":
+  create_workflow()
--- a/py/kubeflow/testing/ci/kf_unittests.py
+++ b/py/kubeflow/testing/ci/kf_unittests.py
@@ -1,13 +1,12 @@
-"""Test Argo workflow"""
+"""Uses an Argo example to generate to create a workflow"""
 
 import requests
 import yaml
 
-def create_workflow():
+def create_workflow(**kwargs):
   """ Loads Argo example YAML and returns dictionary object """
   # TODO: define workflow to run unittests
-  argo_hello_world = requests.get("https://raw.githubusercontent.com/argoproj/"
-  "argo/master/examples/hello-world.yaml")
+  argo_hello_world = requests.get(''.join([v for k, v in kwargs.items()]))
   yaml_result = yaml.safe_load(argo_hello_world.text)
   return yaml_result
 
--- a/py/kubeflow/testing/ci/kf_unittests.py
+++ b/py/kubeflow/testing/ci/kf_unittests.py
@@ -1,4 +1,13 @@
-"""Uses an Argo example to generate to create a workflow"""
+"""
+A smoke test to ensure run_e2e_workflow.py can properly trigger workflows using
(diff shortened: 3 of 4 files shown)


py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -228,56 +232,104 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
...
+
+      ui_url = ("http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/{0}"
               "?tab=workflow".format(workflow_name))
-    ui_urls[workflow_name] = ui_url
-    logging.info("URL for workflow: %s", ui_url)
+      ui_urls[workflow_name] = ui_url
+      logging.info("URL for workflow: %s", ui_url)
 
   # We delay creating started.json until we know the Argo workflow URLs
   create_started_file(args.bucket, ui_urls)
```
kkasravi: it looks like created_started_file is called here and on line 403

py/kubeflow/testing/ci/kf_unittests.py
```diff
@@ -0,0 +1,15 @@
+"""Test Argo workflow"""
+
+import requests
+import yaml
+
+def create_workflow():
```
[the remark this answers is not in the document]
scottilee: But in this example what would I do with the kwargs?

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -17,9 +17,12 @@
...
       tensorflow/*
 
-  - name: lint
-    app_dir: kubeflow/kubeflow/testing/workflows
-    component: workflows
+  - name: workflow-test
+    job_types:
+      presubmit
+    py_func: my_test_package.my_test_module.my_test_workflow
+    kw_args:
```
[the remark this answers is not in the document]
scottilee: Done.

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -73,6 +89,15 @@ def __init__(self, name, app_dir, component, job_types, include_dirs, params):
     self.include_dirs = include_dirs
     self.params = params
 
+class WorkflowPyComponent(object):
+  """Datastructure to represent a Python function to submit a workflow."""
+
+  def __init__(self, name, job_types, py_func, kw_args):
```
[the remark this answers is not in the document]
scottilee: I'll probably just leave this as-is.

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -228,56 +232,104 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
...
+    # check if ks workflow and run
+    if hasattr(w, "app_dir"):
+      ks_cmd = ks_util.get_ksonnet_cmd(w.app_dir)
+
+      # Print ksonnet version
+      util.run([ks_cmd, "version"])
+
+      # If we are scoping this workflow to specific directories, check if any files
+      # modified match the specified regex patterns.
+      dir_modified = False
```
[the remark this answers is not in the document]
scottilee: The only problem is that the Python-defined workflows don't have the `app_dir` property, did you want that property added to them?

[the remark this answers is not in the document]
jlewi: I think I'm suggesting the code should look something like the following

if skip_workflow
   continue

if is_ksonnet_workflow
   ...
elif is_python_workflow

In other words all of the logic to decide whether the workflow should be skipped or not should be run regardless of what type of workflow is that.

If IIUC we only check app_dir to determine whether it is a ksonnet workflow or not. So that part should come later.

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -175,44 +204,19 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
...
-    # Print ksonnet version
-    util.run([ks_cmd, "version"])
 
     # Skip this workflow if it is scoped to a different job type.
     if w.job_types and not job_type in w.job_types:
       logging.info("Skipping workflow %s because job type %s is not one of "
                    "%s.", w.name, job_type, w.job_types)
       continue
 
-    # If we are scoping this workflow to specific directories, check if any files
```
jlewi: I think you want to leave lines 195-215 here. So that we always perform this check to decide whether or not to skip the test regardless of whether its a ksonnet or pyfunc workflow.

py/kubeflow/testing/ci/kf_unittests.py
```diff
@@ -0,0 +1,15 @@
+"""Test Argo workflow"""
```
jlewi: Can you expand the comment please.

Most people will be a bit confused by this file because the filename says unittests but this doesn't look like a unittest file.

py/kubeflow/testing/run_e2e_workflow.py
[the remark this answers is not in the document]
jlewi: See my other comment about lines 195-215

py/kubeflow/testing/ci/kf_unittests.py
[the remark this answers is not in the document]
jinchihe: Maybe pass the helllo_world workflow yaml url by kwaygs as example?

jlewi:
@scottilee Think you might be able to take another look at this tomorrow?

scottilee:
@jlewi Committed some changes and I think it worked? Can you confirm? Log is here in case it gets wiped by the time you see it: `gs://kubernetes-jenkins/pr-logs/pull/kubeflow_testing/451/kubeflow-testing-presubmit/1169481803812048896/build-log.txt`

README.md
```diff
@@ -758,6 +760,32 @@ Follow these steps to add a new test to a repository.
...
+
+       ```
+       workflows:
+        - name: workflow-test
+          py_func: my_test_package.my_test_module.my_test_workflow
+          kwargs:
+              arg1: argument
+       ```
+
+       * **py_func**: Is the Python method to invoke Argo workflows
```
jlewi: nit: "invoke" > to create a python object representing the Argo workflow resource.

prow_config.yaml
```diff
@@ -4,3 +4,20 @@ workflows:
   - app_dir: kubeflow/testing/workflows
     component: workflows
     name: unittests
+
+  - app_dir: kubeflow/kubeflow/testing/workflows
```
jlewi: I assume this is what you want to delete?

prow_config.yaml
```diff
@@ -4,3 +4,20 @@ workflows:
...
+    job_types:
+    - periodic
+    include_dirs:
+    - bootstrap/*
+    - kubeflow/*
+    - testing/*
+    params:
+      workflowName: skipmeplease
+
+  - py_func: kubeflow.testing.ci.kf_unittests.create_workflow
```
jlewi: nit: Can you add a TODO: explaining what this function is doing? i.e. that it is a smoke test to verify that the workflow is actually triggered.

py/kubeflow/testing/ci/kf_unittests.py
```diff
@@ -0,0 +1,14 @@
+"""Uses an Argo example to generate to create a workflow"""
```
jlewi: nit:

"""A smoke test to ensure run_e2e_workflow.py can properly trigger workflows using a py_func.

This module defines a simple py_func which will return the Argo workflow resulting from loading the specified
YAML file.

The purpose of this is to allow the presubmits/postsubmits to verify that run_e2e_workflow.py is actually
triggering workflows using py_func.
"""

prow_config.yaml
jlewi: I think if you wanted to keep this you could. Its not a bad idea to leave this here to verify that presubmits skip files if they aren't changed.

Just add a comment explaining what this is for.  And change app_dir to   some/nonexistent/appdir

In the comment explain that this verifies that workflow is skipped because the test would fail since it doesn't exist.

prow_config.yaml
```diff
@@ -4,3 +4,20 @@ workflows:
   - app_dir: kubeflow/testing/workflows
     component: workflows
     name: unittests
+
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: click_deploy_test
+    name: skipmeplease
+    job_types:
+    - periodic
```
jlewi: This should be presubmit.
If you set it to periodic then you aren't testing whether its being skipped because no matching files were modified.

prow_config.yaml
```diff
@@ -4,3 +4,20 @@ workflows:
...
+    include_dirs:
+    - bootstrap/*
```
jlewi: Have a single include_dirs and set it to something like
- some/nonexistent/dir

This way its clear we don't expect any files to match so tit will always be skipped in presubmit.

Once you make the above changes you could do the same thing but with a py_func and point it to some nonexistent py_func.

This way we also verify skipping is applied to py_funcs.

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -73,6 +89,15 @@ def __init__(self, name, app_dir, component, job_types, include_dirs, params):
     self.include_dirs = include_dirs
     self.params = params
 
+class WorkflowPyComponent(object):
```
jlewi: Why doesn't PyComponent have include_dirs? We want to apply skipping even in the context of PyFunc.

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -62,7 +71,14 @@ def get_namespace(args):
...
 
-class WorkflowComponent(object):
+# imports py_func
+def py_func_import(py_func, kwargs):
+  path, module = py_func.rsplit('.', 1)
+  mod = importlib.import_module(path)
+  met = getattr(mod, module)
+  return met(**kwargs)
+
+class WorkflowKSComponent(object):
```
jlewi: Can we simplify this code by having a single class:

````
class WorkflowComponent(object):
   def  __init__(self, root_dir, data):
       ...
```

Where data is the dictionary loaded from the YAML file.

It doesn't look like there is much difference in the handling of PyWorkflows and KSWorkflows.
It looks like the only difference is how we initialize certain values; e.g. app_dir.

It makes sense to move this processing into the init function.

Using a single class will help eliminate code duplication which should avoid problems caused by slightly different implementations.

You can then define appropriate helper methods e.g.

```
  @property 
  def include_dirs:
     return self.data.get("include_dirs", [])
```

This would simplify the code below because you can remove the has_attr("include_dirs") checks and just do w.include_dirs

README.md
scottilee: Fixed.

py/kubeflow/testing/ci/kf_unittests.py
scottilee: Fixed.

py/kubeflow/testing/run_e2e_workflow.py
scottilee: Fixed.

py/kubeflow/testing/ci/kf_unittests.py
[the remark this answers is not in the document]
scottilee: Fixed.

py/kubeflow/testing/run_e2e_workflow.py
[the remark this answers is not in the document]
scottilee: Fixed.

prow_config.yaml
scottilee: Done.

prow_config.yaml
scottilee: Fixed.

py/kubeflow/testing/ci/kf_unittests.py
scottilee: Added.

prow_config.yaml
scottilee: Fixed.

prow_config.yaml
scottilee: Done.

jlewi:
Let me know when this is ready for another look.

py/kubeflow/testing/run_e2e_workflow.py
scottilee: If we do simplify both classes into a single one I imagine it would look something like this (I think instead of creating the helper methods we can just use default values), do you agree?:

```
class WorkflowComponent(object):
  """Datastructure to represent either a ksonnet component or Python function to
  submit a workflow."""
  def __init__(self, name, app_dir, component, job_types = [], include_dirs = [], params = {}, py_func, kwargs = {}):
    self.name = name
    self.app_dir = app_dir
    self.component = component
    self.job_types = job_types
    self.include_dirs = include_dirs
    self.params = params
    self.py_func = py_func
    self.args = kwargs
```

jlewi: The data parameter is the dict that is loaded from the yaml. The result of loading the yaml is a dictionary with field workflows which is a list of dictionaries so we can just do

```
config = yaml.load(<path to file>)

workflows = []
for w in config["workflows"]
  workflows.append(WorkflowComponent(w))
```
So I think it makes more sense to do the parsing of that dict inside the init function rather than outside the dict.

This way we only have to write the parsing logic once even if we were to construct workflow objects from the YAML in multiple places.

Also IIRC using objects as default values is not a good idea in python because python passes everything by reference. So each instance of WorkflowComponent would end up having self.job_types pointing to the same list. Which is not what we want.

jlewi: If we pass in the dict then we could just do

```
def __init__(self, data):
   self.job_types = data.get("job_types", [])
```

py/kubeflow/testing/run_e2e_workflow.py
scottilee: Fixed.

scottilee:
@jlewi Can you take another look?

jlewi:
Thanks @derekhh!

/lgtm
/approve

PR #454, opened 2019-09-03
[WIP] Add subdirectories to JUnit artifacts

### LINKED ISSUE #372: junit files from multiple runs of a workflow with different parameters clobber each other
Right now we run our E2E test workflows with different parameters (e.g. the kfctl_e2e_workflow).

But the junit files are often times not unique. As a result we end up clobbering the runs.

Example:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow_kubeflow/kubeflow-postsubmit/1121841104535162880/

We have two different workflows
/mnt/test-data-volume/kubeflow-postsubmit-kfctl-go-iap-97ce5be-2880-8b97/output/artifacts/junit_kfctl-go-delete-test.xml 
/mnt/test-data-volume/kubeflow-postsubmit-kfctl-go-iap-istio-97ce5be-2880-bc9c/output/artifacts/junit_kfctl-go-delete-test.xml

They would both be uploaded to the same location. 

Except in this case copy-artifacts never ran.
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-postsubmit-kfctl-go-iap-istio-97ce5be-2880-bc9c?tab=workflow&nodeId=kubeflow-postsubmit-kfctl-go-iap-istio-97ce5be-2880-bc9c-605660832

/cc @zabbasi

derekhh:
**Background**

This PR is related to #372.

**Description**

Following @jlewi's suggestion and @zabbasi's PR #3381, in this PR I've manually updated the artifact directory for the testing workflows to `junit/workflow_name`. This means the final junit XML file will be uploaded to the location `.../artifacts/junit/workflow_name/junit_{test_name}.xml`.

**Tests**

- [x] Verified that the artifacts are uploaded to specific subdirectories under the 'junit' folder and that these new XML locations can still correctly trigger JUnit viewers in Spyglass.

    Artifact Link: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/454/kubeflow-testing-presubmit/1168745601043533826/artifacts/junit/kubeflow-testing-presubmit-unittests-454-c462c96-3826-8194/

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/454)


derekhh:
@jlewi - The workflow is still in Pending state in Spyglass (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/454/kubeflow-testing-presubmit/1168745601043533826/) but both build logs and the Argo link (http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-testing-presubmit-unittests-454-c462c96-3826-8194?tab=timeline&nodeId=null) show that the workflow completed successfully.

I'm seeing the same issue on @kkasravi's test (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/446/kubeflow-testing-presubmit/1168206423964782599/). Do you know if this is a known recent regression?

jlewi:
@derekhh thanks for picking this up.

For an issue lets not close the issue until we have verified that things are working correctly after the PR  is merged. So after the PR is merged we should look at a post submit run of some test like kfctl and verify that all the results are in the proper location. So to prevent GitHub from auto-closing the issue we should avoid putting "fix #..." in the PR description. instead we should say something like "related to #"...

jlewi:
Also it looks like this PR would only make the change for the tests triggered by kubeflow/testing. We will need to make the same change for the E2E tests in kubeflow/kubeflow.

Not sure modifying the ".cache" directories has any effect.

jlewi:
I don't know why the tests failed. It looks like the workflow succeeded and the job succeeded. 
Maybe it was related to the problems caused by the changes to run_e2e_workflow.py that have since been rolled back?

derekhh:
@jlewi 

* Thanks for the comment. I was not aware "fix" in the PR description would automatically trigger the referred issue. TIL. :) I updated the PR description.

* I also reverted my changes to the `.cache` folder. I'll open a PR in kubeflow/kubeflow after we verify this PR works.

jlewi:
@derekhh Did you revert too much? It looks like this PR is now basically empty? Do you need to add "junit/<workflow name>"  or something like that as a subdirectory of artifacts?

derekhh:
@jlewi - Oh that was an intentional choice to see whether the Prow test failure was actually a result of my PR. Turns out it actually seems to be so I'm figuring out what's going on. :p

PR #463, opened 2019-09-17
Use a py_func and not ksonnet to define the unittest E2E workflow; update kf_unittests to create a valid workflow

jlewi:
* Create a simple CLI tool e2e_tool.py to support working with py_funcs
  for E2E tests

* run_e2e_workflow.py should pass the name and namespace to the py_func.

* Add to util.load_kube_config an option to skip printing out the kube
  config.

* Remove the ksonnet workflow for E2E tests.

* Create argo_build_util to include some reusable utilities for building workflows

* Add an option to prow_config to list python directories to add to the python path in order to
  import the py_func

Related to kubeflow/kubeflow#3035 migrate e2e tests off ksonnet

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/463)


py/kubeflow/testing/ci/kf_unittests.py
```diff
@@ -1,23 +1,248 @@
...
+       'value': os.path.join(self.test_dir, 'kfctl_test/.kube/kubeconfig')},
+    ]
+
+    task_template["container"]["env"].extend(common_env)
+
+    task_template = argo_build_util.add_prow_env(task_template)
+
+    return task_template
+
+  def build(self):
```
zhenghuiwang: If I have to create an argo workflow for another repository, seems I have to copy this unittest.py and modify only these specific tasks.

Is it better to move all these functions to argo_build_util.py; and have a smaller interface for user to use, i.e. only `newTask(name, cmd, deps)` and `newWorkflow(tasks, entry)`?

jlewi: @zhenghuiwang We can iterate over time. If you look at kubeflow/kubeflow#4148 which defines the E2E test for Argo workflow. I think there will be more variability between different E2E tests. So I'm not sure what the right abstraction/syntactic sugar is.

I think after we write a few more pipelines we will get more insight and can refine it.

PR #466, opened 2019-09-20
Fix some issues with run_e2e_workflow.py and pyfunc

jlewi:
* When trying to import the py_func in the E2E tests for kubeflow/kubeflow repository, I ran into problems with the import failing because the module kubeflow.kubeflow could not be found

  * The issue is tracked kubeflow/testing#467

  * Essentially dynamically  adding the kubeflow/kubeflow/py (i.e. from inside python) didn't work
     * Could be an issue with namespace packaging

* I spent hours trying to make dynamically importing the modules work but I couldn't figure it out

* Setting python path before invoking python appears to work. So as a workaround we have run_e2e_workflow.py shell out to e2e_tool which writes the workflow to a file. run_e2e_workflow.py then loads it from the file.

* Create a K8s job to run run_e2e_workflow.py to simulate what happens
  under prow

* Related to kubeflow/kubeflow#3035 migrate e2e tests off ksonnet

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/466)

### DIFF
--- a/hack/run_e2e_workflow_k8s_job.yaml
+++ b/hack/run_e2e_workflow_k8s_job.yaml
@@ -1,0 +1,100 @@
+# A Sample K8s job for running run_e2e_workflow.py
+#
+# This is useful for debugging/testing run_e2e_workflow.py during development.
+# The job is intended to simulate what runs in our prow job.
+# https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-presubmits.yaml
+# Using a K8s job gives us more control over what we run. For example
+# we can use EXTRA_REPOS to check out a PR of kubeflow/testing to checks changes to run_e2e_workflow.py
+apiVersion: batch/v1
+kind: Job
+metadata:
+  generateName: run-e2e-workflow
+  namespace: kubeflow-test-infra
+  labels:
+    app: run-e2e-workflow
+spec:
+  template:
+    spec:
+      initContainers:
+        # This init container checks out the source code.
+        - command:
+          - /usr/local/bin/checkout_repos.sh
+          # TODO(jlewi): Why are we checkout out fairing?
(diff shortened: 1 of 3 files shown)


yanniszark:
@jlewi thanks a lot for the detailed explanation.
I don't know why the dynamic change of the sys.path doesn't work for the namespaced package.
However, since PYTHONPATH seems to work, why not do:

```bash
PYTHONPATH="/src/${REPO_OWNER}/${REPO_NAME}/py:$PYTHONPATH"
```

before we call `run_e2e_workflow.py`, inside the `run_workflows.sh` file?

Seems simpler than shelling out again inside Python.

jlewi:
@yanniszark I agree with you one caveat. To get pythonpath we need to load and parse the YAML file. I'm not really a fan of doing that level of complex processing in bash. I think its more readable to do it in python. 

Either way we end up in a world where we have a shim script that sets the python path and then invokes the main python program.

What's really hacky about the current implementation is we are splitting the python processing across two subprocesses. I think it would be cleaner to do make the first python script dirt simple e.g.

````
if __name__ == "___main__"::
    load prowconfig
    get python paths
    subprocess.check_call(process_prowconfig)
````

I considered doing that refactoring now but decided to leave it for later since this appears to work and I think there's more urgency to getting a fix in that unblocks other repos from using py_func.

**Comment by yanniszark** [comment]
> To get pythonpath we need to load and parse the YAML file.

Maybe I'm missing something here.
I was talking about a simple `export` inside the `run_workflows.sh` file.

```bash
export PYTHONPATH="/src/${REPO_OWNER}/${REPO_NAME}/py:$PYTHONPATH"
```

All env vars should be available so not sure why we need to parse YAML.

In any case, I want to unblock your work regarding using the py_func and this seems to do the trick.

Issue #461, opened 2019-09-16
Kubeflow auto-deployments from master endpoint not accessible; no SSL certificate


jlewi:
The kubeflow auto deployments from master are not accessible via IAP.

It looks like the certificate isn't getting created (probably as a result of switching to managed certificates in 0.7).

Here are the events for one of the master clusters

```
  Type     Reason                 Age                    From                     Message
  ----     ------                 ----                   ----                     -------
  Warning  Sync                   13m (x132 over 12h)    loadbalancer-controller  Error during sync: error running load balancer syncing routine: loadbalancer istio-system-envoy-ingress--44438c1b268ad23e does not exist: pre-shared-cert errors: errors - mcrt-d37050b4-d08c-4394-bde9-ca24850fffa4: googleapi: Error 404: The resource 'projects/kubeflow-ci-deployment/global/sslCertificates/mcrt-d37050b4-d08c-4394-bde9-ca24850fffa4' was not found, notFound
  Normal   SslCertificateMissing  3m13s (x134 over 12h)  loadbalancer-controller  googleapi: Error 404: The resource 'projects/kubeflow-ci-deployment/global/sslCertificates/mcrt-d37050b4-d08c-4394-bde9-ca24850fffa4' was not found, notFound
```

Here are the events for the managed certificate.

```
jlewi@jlewi-carbon-glaptop:~/git_kubeflow-manifests$ kubectl -n istio-system describe managedcertificate  gke-certificate
Name:         gke-certificate
Namespace:    istio-system
Labels:       kustomize.component=iap-ingress
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"networking.gke.io/v1beta1","kind":"ManagedCertificate","metadata":{"annotations":{},"labels":{"kustomize.component":"iap-in...
API Version:  networking.gke.io/v1beta1
Kind:         ManagedCertificate
Metadata:
  Creation Timestamp:  2019-09-16T00:17:13Z
  Generation:          1
  Resource Version:    22805
  Self Link:           /apis/networking.gke.io/v1beta1/namespaces/istio-system/managedcertificates/gke-certificate
  UID:                 55636b26-d817-11e9-a9ab-42010a8e015d
Spec:
  Domains:
    kf-vmaster-n02.endpoints.kubeflow-ci-deployment.cloud.goog
Status:
  Certificate Name:  mcrt-d37050b4-d08c-4394-bde9-ca24850fffa4
  Domain Status:
Events:
[...]

PR #474, opened 2019-09-27
Improve cleanup of SSl certificates

jlewi:
* Parse the SSL certificates so that we get the associated domain.

  Use this to expire SSL certificates from pre/postsubmits after a couple
  hours while leaving the certificates for the auto deployments around longer.

* Minor fixes to some of the other tools.

Related to #461

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/474)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -14,6 +14,9 @@ import traceback
 import tempfile
 import yaml
 
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+
 from kubeflow.testing import argo_client
 from kubeflow.testing import util
 from kubernetes import client as k8s_client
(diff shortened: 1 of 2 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -608,9 +611,30 @@ def cleanup_certificates(args):
...
+      if not "certificate" in d:
+        logging.warning("Certificate %s is missing certificate", name)
+        continue
+
+      raw_certificate = d["certificate"]
+
+      cert = x509.load_pem_x509_certificate(raw_certificate.encode('utf-8'), default_backend())
+
+      # TODO(jlewi): Is there a way to do this without accessing protected attributes
+      domain = cert.subject._attributes[0]._attributes[0].value # pylint: disable=protected-access
```
lluunn: Are two _attributes[0] intended?
What's the format of subject?

jlewi: Yeah its a bit of a hack I couldn't figure out what the proper way to do this was which is why I'm accessing private attributes.
It ends up being the domain e.g jlewi-0-6-2-rc.endpoints.jlewi-dev.cloud.goog

PR #477, opened 2019-10-07
fix wait_for_deployment to also check for apps/v1 deployments

yanniszark:
/cc @jlewi @swiftdiaries @kkasravi 

Signed-off-by: Yannis Zarkadas <yanniszark@arrikto.com>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/477)

### DIFF
--- a/py/kubeflow/testing/util.py
+++ b/py/kubeflow/testing/util.py
@@ -410,9 +410,13 @@ def wait_for_deployment(api_client,
     minutes=timeout_minutes)
 
   ext_client = k8s_client.ExtensionsV1beta1Api(api_client)
+  ext_client_apps = k8s_client.AppsV1beta1Api(api_client)
 
   while datetime.datetime.now() < end_time:
-    deploy = ext_client.read_namespaced_deployment(name, namespace)
+    try:
+      deploy = ext_client.read_namespaced_deployment(name, namespace)
+    except rest.ApiException:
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/util.py
```diff
@@ -410,9 +410,13 @@ def wait_for_deployment(api_client,
     minutes=timeout_minutes)
 
   ext_client = k8s_client.ExtensionsV1beta1Api(api_client)
+  ext_client_apps = k8s_client.AppsV1beta1Api(api_client)
 
   while datetime.datetime.now() < end_time:
-    deploy = ext_client.read_namespaced_deployment(name, namespace)
+    try:
+      deploy = ext_client.read_namespaced_deployment(name, namespace)
+    except rest.ApiException:
```
jlewi: Do we need this? Even if an object is created as an extension/v1beta1 is it readable using the v1 API?
If not can we add a TODO to track cleaning this up after upgrading all our manifests to v1?

yanniszark: After looking around a bit more, it seems that we don't need it.
The issue was with the new changes to kfctl (didn't default to kustomize if the packageManager field was empty).
See: https://github.com/kubeflow/kubeflow/pull/4154/commits/8f7893e910f965f449954fb97db57d7315464b4c

Issue #480, opened 2019-10-14
kubeflow-ci-deployment out of private address space on default network


jlewi:
Starting to see this area again.

#476 was the previous instance of this issue.

It does look like we are leaking some clusters

```
gcloud --project=kubeflow-ci-deployment container clusters list --format="table(name, location, status, createTime)" --sort-by=createTime
NAME            LOCATION       STATUS        CREATE_TIME
deployapp       us-east1-d     RUNNING       2019-04-26T22:35:53+00:00
apps            us-central1-a  RUNNING       2019-09-18T23:04:32+00:00
kf-vmaster-n04  us-east1-b     RUNNING       2019-10-02T12:13:19+00:00
kf-vmaster-n00  us-east1-b     RUNNING       2019-10-03T00:09:58+00:00
myapp2          us-central1-a  RUNNING       2019-10-03T04:18:04+00:00
kf-vmaster-n01  us-east1-b     RUNNING       2019-10-03T12:12:28+00:00
kf-vmaster-n02  us-east1-b     RUNNING       2019-10-04T00:12:04+00:00
kfctl-e899      us-central1-a  RUNNING       2019-10-09T05:50:58+00:00
kfctl-f766      us-central1-a  RUNNING       2019-10-10T01:57:22+00:00
kfctl-d643      us-central1-a  RUNNING       2019-10-10T23:53:57+00:00
kfctl-f282      us-central1-a  RUNNING       2019-10-11T23:19:33+00:00
kfctl-d448      us-central1-a  RUNNING       2019-10-11T23:28:00+00:00
kfctl-f5dc      us-central1-a  RUNNING       2019-10-12T01:55:03+00:00
kfctl-e1d5      us-central1-a  RUNNING       2019-10-12T06:56:23+00:00
kf-v0-6-n04     us-east1-b     RUNNING       2019-10-12T12:30:23+00:00
kfctl-d472      us-central1-a  RUNNING       2019-10-12T23:23:27+00:00
kf-v0-6-n00     us-east1-b     RUNNING       2019-10-13T00:21:04+00:00
kf-v0-6-n01     us-east1-b     RUNNING       2019-10-13T12:15:05+00:00
kf-v0-6-n02     us-east1-b     RUNNING       2019-10-14T00:15:11+00:00
kfctl-d256      us-central1-a  RUNNING       2019-10-14T07:48:46+00:00
kf-v0-6-n03     us-east1-b     RUNNING       2019-10-14T12:16:20+00:00
[...]

PR #485, opened 2019-10-15
Aggressively GC clusters that are in an error state.

jlewi:
Related to #480.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/485)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -881,6 +881,8 @@ def cleanup_clusters(args):
 
   expired = []
   unexpired = []
+  stopping = []
+
   for zone in args.zones.split(","):
     clusters = clusters_client.list(projectId=args.project, zone=zone).execute()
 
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -1088,6 +1103,19 @@ def main():
 
   add_deployments_args(parser_deployments)
   parser_deployments.set_defaults(func=cleanup_deployments)
+
+  ######################################################
+  # Parser for clusters
+  parser_clusters = subparsers.add_parser(
+    "clusters", help="Cleanup clusters")
+
+  parser_clusters.add_argument(
```
lluunn: we have this already?
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/cleanup_ci.py#L970

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -881,6 +881,8 @@ def cleanup_clusters(args):
 
   expired = []
   unexpired = []
+  stopping = []
```
gabrielwen: could you help correct the log in LN 875?

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -904,7 +906,19 @@ def cleanup_clusters(args):
       insert_time_utc = insert_time + datetime.timedelta(hours=-1 * hours_offset)
       age = datetime.datetime.utcnow()- insert_time_utc
 
-      if age > datetime.timedelta(hours=args.max_age_hours):
+      if c.get("status", "") == "ERROR":
```
gabrielwen: should we also GC `degraded` clusters?
https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.Status

py/kubeflow/testing/cleanup_ci.py
jlewi: Done

py/kubeflow/testing/cleanup_ci.py
jlewi: Done.

py/kubeflow/testing/cleanup_ci.py
jlewi: That's adding subparser commands and is only for the deployments command. Here we are are doing it for the clusters command.

PR #487, opened 2019-10-16
Check environment variable BUILD_ID and use that for the prow bucket.

jlewi:
* See kubeflow/kfctl#58 junit artifacts aren't being uploaded for some of our
  tests because the GCS bucket is being set incorrectly.

* It looks like the problem is that our Argo workflows are setting BUILD_ID
  and not BUILD_NUMBER

* To avoid breaking workflows that might still be relying on BUILD_NUMBER
  we try BUILD_ID and fall back to BUILD_NUMBER

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/487)

### DIFF
--- a/py/kubeflow/testing/prow_artifacts.py
+++ b/py/kubeflow/testing/prow_artifacts.py
@@ -111,6 +111,18 @@ def get_gcs_dir(bucket):
   job_name = os.getenv("JOB_NAME")
   job_type = os.getenv("JOB_TYPE")
 
+  # Based on the prow docs the variable is BUILD_ID
+  # https://github.com/kubernetes/test-infra/blob/45246b09ed105698aa8fb928b7736d14480def29/prow/jobs.md#job-environment-variables
+  # But it looks like the original version of this code was using BUILD_NUMBER.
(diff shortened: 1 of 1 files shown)


jlewi:
prow_artifacts ran the following copy command

```
gsutil -m rsync -r /mnt/test-data-volume/kubeflow-testing-presubmit-py-unittests-487-dcec6d2-1284-d306/output gs://kubeflow-ci_temp/pr-
logs/pull/kubeflow_testing/487/kubeflow-testing-presubmit/1184598609765601284
```

The directory in spyglass is
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/487/kubeflow-testing-presubmit/1184598609765601284/

So the bucket is wrong. I should have verified before requesting a review.

[kubeflow-testing-presubmit-py-unittests-487-dcec6d2-1284-d306-3656271867.log.txt](https://github.com/kubeflow/testing/files/3736885/kubeflow-testing-presubmit-py-unittests-487-dcec6d2-1284-d306-3656271867.log.txt)

PR #488, opened 2019-10-17
Fix bucket where artifacts are uploaded to

jlewi:
* Prow spyglass are still using kubernetes-jenkins; if we use kubeflow-ci_temp
  the results won't show up in spyglass.

* Also we don't need to set suffix to fakesuffix.

* related to kubeflow/kfctl#58

* Fix a bunch of lint errors
  * since junit results from lint weren't being surfaced in prow a bunch of issues crept in over time.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/488)

### DIFF
--- a/py/kubeflow/testing/run_e2e_workflow.py
+++ b/py/kubeflow/testing/run_e2e_workflow.py
@@ -87,7 +87,7 @@ def py_func_import(py_func, kwargs):
   met = getattr(mod, create_function)
   return met(**kwargs)
 
-class WorkflowComponent(object): # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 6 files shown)


py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -197,7 +197,7 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
     extra_py_paths.append(path)
 
   kf_test_path = os.path.join(args.repos_dir, "kubeflow/testing/py")
-  if not kf_test_path in extra_py_paths:
```
abhi-g: heh!

Issue #489, opened 2019-10-17
Support grouping in testgrid


jlewi:
Our E2E tests run Argo workflows.  We would like to group related tests when viewing the results in testgrid. For example we'd like to see all tests corresponding to the "GCP-IAP" config grouped together vs the same tests running in the "GCP-basic auth" config.

Test grid supports organizing by test target which corresponds to classname in the junit xml file.

When generating junit files from pytest we can use record_my_xml_attribute
http://doc.pytest.org/en/latest/usage.html#record-xml-attribute

to override the classname (it defaults to python module name).

The obvious value to use for the class name is the name assigned in the prow_config file
https://github.com/kubeflow/kubeflow/blob/70d3119fc5f075ff9cf715e6bfe0a8ce00e05736/prow_config.yaml#L32

This ensures each set of parameters would correspond to a different grouping.

To support that we should have run_e2e_worfklow.py pass that name as argument "junit_class_name" to all create_worfklow functions.
https://github.com/kubeflow/testing/blob/c6299257d8ff944a6b0b25b10abf97bff8c955e4/py/kubeflow/testing/run_e2e_workflow.py#L330

Then a workflow can set an environment variable to that value and pytest functions can use that environment variable to override the class name.

PR #490, opened 2019-10-17
Support subdirectories for junit files and junit_class_name

jlewi:
Support subdirectories for junit files and grouping in test grid by name
  
  * We want to support putting junit files in subdirectories e.g.
    junit/junit_workflow1/junit_test1.xml
    junit/junit_workflow2/junit_test1.xml
  
    * prow_artifacts.py needs to fetch the junit files from GCS in order to
      determine if there were any test failures to properly set the prow job
      status
  
    * The code however was making an assumption that all results were in
      the artifacts dir and not subdirectories
  
  * run_e2e_workflow.py should set a keyword argument test_target_name
    to the name of the workflow (as set in prow_config.yaml) for each of
    the python functions creating argo workflows.
  
    * We want to use the junit attribute classname to group tests coming
      from the same workflow. To do that we need to know the name of the workflow.
  
    * We also use this as a prefix for the test name to support grouping
      hierarchically
  
  * Related to kubeflow/testing#489
  
  * Update kfunittests to Use a junit subdirectory to make sure subdirectories are properly handled.
  
  * Update docs
  
  * update toc.
  
  * Convert test_py_lint to use pytest. Do this because we want to be able to test the use of junit features.
  
  * Add a separate mkdir step.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/490)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@
     - [Argo UI](#argo-ui)
     - [Stackdriver logs](#stackdriver-logs)
   - [Debugging Failed Tests](#debugging-failed-tests)
+    - [Logs and Cluster Access for Kubeflow CI](#logs-and-cluster-access-for-kubeflow-ci)
+      - [Access Control](#access-control)
     - [No results show up in Gubernator](#no-results-show-up-in-gubernator)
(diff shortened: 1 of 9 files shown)


jlewi:
On the latest run the name for the pylint test was

```
py-unittests/test_py_lint
```

So overriding the test name in the junit appears to work. No results are showing up in the test grid at all so I can't verify that's working. That seems unrelated to this PR.

Issue #471, opened 2019-09-27
Kubeflow auto-deployments from master failing; error setting project


jlewi:
Here's the stack trace from the most recent failure

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/mnt/test-data-volume/auto_deploy/auto-deploy-master-1569499200/testing/py/kubeflow/testing/create_kf_instance.py", line 307, in <module>
    main()
  File "/mnt/test-data-volume/auto_deploy/auto-deploy-master-1569499200/testing/py/kubeflow/testing/create_kf_instance.py", line 259, in main
    deploy_with_kfctl_go(kfctl_path, args, app_dir, env)
  File "/mnt/test-data-volume/auto_deploy/auto-deploy-master-1569499200/testing/py/kubeflow/testing/create_kf_instance.py", line 103, in deploy_with_kfctl_go
    config_spec["spec"]["project"] = args.project
KeyError: 'spec'

```

jlewi:
I think the problem is that we are trying to pull the config from kubeflow/kubeflow but the manifest has moved to kubeflow/manifests.

Here's the invocation.

```
        - /usr/local/bin/auto_deploy.sh
        - --repos=kubeflow/kubeflow;kubeflow/testing
        - --project=kubeflow-ci-deployment
        - --job_labels=/etc/pod-info/labels
        - --data_dir=/mnt/test-data-volume/auto_deploy
        - --base_name=kf-vmaster
        - --max_num_cluster=5
        - --zone=us-east1-b
        - --github_token_file=/secret/github-token/github_token
        - --kfctl_config=https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_gcp_iap.yaml

```

jlewi:
Still failing

```
gcloud --project=kubeflow-ci-deployment container clusters list --format="table(name, location, status, createTime)" --sort-by=createTime
NAME            LOCATION       STATUS    CREATE_TIME
deployapp       us-east1-d     RUNNING   2019-04-26T22:35:53+00:00
apps            us-central1-a  RUNNING   2019-09-18T23:04:32+00:00
kf-vmaster-n04  us-east1-b     RUNNING   2019-10-02T12:13:19+00:00
[...]

Issue #444, opened 2019-08-19
Auto-deployments of Kubeflow should use unique names and not recycle names - Upgrade to v0.7


jlewi:
Our auto-deployments currently recycle names; e.g.
kf-vmaster-n??...

We originally did this because with IAP we used to need to create register each hostname manually. With universal redirect that should no longer be necessary.

In the past, reusing hostnames would lead to lets-encrypt quota errors. We worked around this by generating self-signed certificates.

But on master we have now switched to using GKE managed certificates and uploading our own self-signed certificate no longer works.

It would be nice if GKE managed certificates avoids the quota issues if we reuse an endpoint.

However, we should now be able to use unique hostnames for each auto-deployment. That should simplify the auto-deployment logic.

To use unique names e.g.

kf-v0-6-%YYYY%MM%DD-%HH%MM%SS

We will need to fix our cleanup logic to delete older deployments and only keep the N most recent.

We can reuse our existing cleanup_ci.py script and just modify that.

jlewi:
The cron job is now failing.

```
INFO|2019-10-25T00:12:40|/src/kubeflow/testing/py/kubeflow/testing/util.py|71| Error:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubeflow.error): Code 500 with message: gcp apply could not update deployment manager Error could not update storage-kubeflow.yaml; Insert deployment error: googleapi: Error 400: Invalid value for field 'resource.labels': ''.  Label key 'GIT_LABEL' violates format constraints. The key must start with a lowercase character, can only contain lowercase letters, numeric characters, underscores and dashes. The key can be at most 63 characters long. International characters are allowed., invalid
```

jlewi:
We need to update 
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/get_kf_testing_cluster.py

To work with the new naming pattern.

jlewi:
We need to update cleanup_ci.py - the max duration of clusters is shorter then the frequency with which we are auto-deploying clusters from master

jlewi:
This should be fixed.

PR #495, opened 2019-10-18
Auto deploy job needs to use the new kfctl syntax.

jlewi:
Auto deploy job needs to use the new kfctl syntax; also use unique names

Related to #471

* Don't set name in the spec because we want to infer it form directory.

* Create a new script to deploy with a unique name

* Related to: https://github.com/kubeflow/testing/issues/444

* Update cleanup script to clean up new auto-deployed clusters

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/495)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -17,19 +17,26 @@ from kubernetes import client as k8s_client
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
 
+# See https://github.com/kubeflow/testing/issues/444
+# We are switching to unique names for auto deployments
+# So this matches the new ones.
+AUTO_DEPLOY_PATTERN = re.compile("kf-vmaster-(?!n\d\d)")
+
 # Regexes that select matching deployments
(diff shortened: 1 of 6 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -17,19 +17,26 @@
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
 
+# See https://github.com/kubeflow/testing/issues/444
+# We are switching to unique names for auto deployments
+# So this matches the new ones.
+AUTO_DEPLOY_PATTERN = re.compile("kf-vmaster-(?!n\d\d)")
```
gabrielwen: this only matches auto-deployment from master?  how about current stable release?

jlewi: This is deliberate. The existing auto deployed clusters aren't cleaned up by cleanup_ci.py. Instead the auto deployment script recycles clusters as it cycles through names.

jlewi:
I verified this worked.
kf-vmaster-1023-237.endpoints.kubeflow-ci-deployment.cloud.goog

PR #500, opened 2019-10-23
upgrade kubectl to v1.14.0 in test-worker

jinchihe:
upgrade kubectl to v1.14.0 in test-worker, the currect new V1.10.0 is too old so caused the porblem beflow.
```
error: SchemaError(io.k8s.api.certificates.v1beta1.CertificateSigningRequestList): invalid object doesn't have additional properties
```
See more in https://github.com/kubeflow/examples/pull/668 and https://github.com/kubeflow/examples/issues/665

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/500)


jinchihe:
@jlewi Seems the test-worker image is not refreshed.

PR #511, opened 2019-10-31
Add a method to build checkout repos dictionary

richardsliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/511)

### DIFF
--- a/py/kubeflow/testing/util.py
+++ b/py/kubeflow/testing/util.py
@@ -31,12 +31,6 @@ MASTER_REPO_NAME = "k8s"
 # How long to wait in seconds for requests to the ApiServer
 TIMEOUT = 120
 
-DEFAULT_REPOS = {
-    "kubeflow/kubeflow": "HEAD",
-    "kubeflow/testing": "HEAD",
-    "kubeflow/tf-operator": "HEAD"
-}
-
 def run(command,
         cwd=None,
         env=None,
@@ -104,23 +98,18 @@ def combine_repos(list_of_repos):
 
   Args:
     list_of_repos: A list of repos to checkout, each one in the format of
-      "owner/name@commit".
+      "owner/name@commit". Later values override earlier ones.
   Returns:
-    repos: A dictionary of repository names to commit hashes. Uses DEFAULT_REPOS as a base.
+    repos: A dictionary of repository names to commit hashes.
   """
 
-  # 1. Convert list_of_repos to a dictionary
-  # where key is "repo_owner/repo_name" and value is the
-  # commit hash.
-  user_repos = {}
+  # Convert list_of_repos to a dictionary where key is "repo_owner/repo_name"
+  # and value is the commit hash. By convention, values that appear later in
+  # the list would override earlier ones.
+  repos = {}
   for r in list_of_repos:
     parts = r.split('@')
-    user_repos[parts[0]] = parts[1]
-
-  # 2. Starting with the default repos, merge in any user-defined
-  # repos with commit hashes.
-  repos = DEFAULT_REPOS
-  repos.update(user_repos)
+    repos[parts[0]] = parts[1]
 
   return repos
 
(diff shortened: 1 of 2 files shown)


py/kubeflow/testing/util.py
```diff
@@ -93,6 +99,34 @@ def run(command,
 def run_and_output(*args, **argv):
   return run(*args, **argv)
 
+def build_repos_dict(main_repo, extra_repos):
```
jlewi: Where does main_repo come from? Is it coming from the prow env?

Would it be useful to have a function

```
def prow_var_to_repo_spec(prow_dict)
```

That would take the prow variables and convert it to the repo spec? I find myself writing that code over and over again.  e.g.

https://github.com/kubeflow/kubeflow/blob/41dd8167ff4171d6acf12822d19c571230cf04ce/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L189

Would it be simpler to simplify the function signature to just be

```
def combine_repos(list_of_repos)
```

list_of_repos would then just be a list of repos in the form ${ORG}/{REPO}:version

Combine would then group the repos by {ORG}/{REPO} with the value on the right taking precedence.

So then in the case of kubeflow/kubeflow we could just write

combine_repos(["kubeflow/kubeflow@HEAD", "kubeflow/tf_operator@HEAD", prow_var_to_repo_spec(os.environ)] + extra_repos)

richardsliu: In the latter example, does it mean that the caller is expected to pass in default repos to combine_repos? That would require us to modify more places.

richardsliu: Also, there is https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/argo_build_util.py#L211

So outside of this method, we could do something like
```
repos = combine_repos([argo_build_util.get_repo_from_prow_env()] + extra_repos)
repos_str = ','.join(['%s@%s' % (key, value) for (key, value) in repos.items()])
```

jlewi: > In the latter example, does it mean that the caller is expected to pass in default repos to combine_repos? That would require us to modify more places.

Doesn't the caller have to pass in default repos? The default repos are test specific. e.g.

```
DEFAULT_REPOS = {
    "kubeflow/kubeflow": "HEAD",
    "kubeflow/testing": "HEAD",
    "kubeflow/tf-operator": "HEAD"
}
```

These defaults are specific to the kfctl_e2e_test. But they wouldn't make sense as the defaults for other tests; e.g. some of the E2E tests in kubeflow/examples

py/kubeflow/testing/util.py
```diff
@@ -31,6 +31,12 @@
 # How long to wait in seconds for requests to the ApiServer
 TIMEOUT = 120
 
+DEFAULT_REPOS = {
```
jlewi: These DEFAULTS are really specific to the kfctl_e2e_test.
Lots of E2E tests won't need the tf-operator repo; kfctl only needs it because its running some TFJob tests.

kubeflow/testing is arguably the only repo that would be universally checked out because that's where we are hosting test utilities to be reused across all repos.

py/kubeflow/testing/util.py
```diff
@@ -93,6 +99,30 @@ def run(command,
...
+  # 1. Convert list_of_repos to a dictionary
+  # where key is "repo_owner/repo_name" and value is the
+  # commit hash.
+  user_repos = {}
+  for r in list_of_repos:
+    parts = r.split('@')
+    user_repos[parts[0]] = parts[1]
+
+  # 2. Starting with the default repos, merge in any user-defined
+  # repos with commit hashes.
```
jlewi: Per comment above. I think the caller should pass in defaults because they will be different for each test.
Checking out repos can be expensive so we don't necessarily want to check out repos we don't need.

richardsliu:
@jlewi PTAL

PR #527, opened 2019-11-11
[WIP] Fix bug with getting credentials in bulk_deploy.py

jlewi:
Numerous fixes and improvements to bulk deploying Kubeflow with v06

Fix bug with getting credentials in bulk_deploy

* Need to support obtaining credentials to talk to K8s API servier using
  the pods service accounnt by calling load incluster config.

* With v06 deployments we need to handle the case where the DM API might
  not be enabled yet and enable it before checking if KF exists.

* Don't use workload identity because with workload identity we are seeing
  performance problems when issuing a lot of requests that hit the gke metadata
  server

Improve the ability to continue deploying v06 deployments that failed.

* With v06 applications check if the ingress actually exists and if
  not we will rerun deploy.

  * Delete istio-ingressgateway service if its not fully deployed to try to recover.

* Add a util method for getting Kubernetes credentials

  * When running in a pod there are two cases we need to handle
    * using the KSA to authenticate to the K8s APIServer the pod is running
    * Using a kubeconfig file to talk to a different cluster.

* Add resource requests to the create_unique_kf_instance jobs to better
  handle scheduling of many jobs

* Create code to delete Kubeflow deployments in bulk

* Don't use a CSV file to provide info about what deployments to create
  * Uploading CSV files to GCS turns out to be a lot of friction

  * instead just take command line arguments which specify the range of
    projects to iterate over.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/527)

### DIFF
--- a/codelabs/README.md
+++ b/codelabs/README.md
@@ -6,7 +6,15 @@ of Kubeflow in codelab test projects.
 The scripts are intended to run on a kubeflow cluster.
 
 **project**: kf-codelab-admin
+**cluster**: codelab-admin-v06
+   * This is a Kubeflow v0.6 cluster
+   * This is an attempt to work around the problems we are seeing with workload identity enabled clusters
+**namespace**: kubeflow-jlewi
+    * Use this namespace because this has K8s secrets with the GCP service account credentials
+
 **cluster**: codelab-admin
+   * This is a Kubeflow v0.7 cluster with workload identity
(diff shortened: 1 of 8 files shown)


zhenghuiwang:
there is lint issue, but /lgtm

jlewi:
Need to make some other changes.

jlewi:
/hold cancel

This should be ready to go assuming tests pass

Issue #528, opened 2019-11-25
cleanup_ci doesn't clean up instance groups properly


gabrielwen:
All the instance groups were skipped during GC because infra type couldn't be determined:
https://github.com/kubeflow/testing/blob/25b87a36f03c207f69abbc9e18500e9ac75647d0/py/kubeflow/testing/cleanup_ci.py#L324

the reason is because instance groups are not named as other resources:
- gke-apps-apps-cpu-pool-v1-a3e58689-grp
- gke-kf-vmaster-1124--kf-vmaster-1124--fae5daa0-grp
- gke-myapp2-myapp2-cpu-pool-v1-e8485b6e-grp
- k8s-ig--00bfab21c3a055cc

jlewi:
@gabrielwen Which projects do the above instance groups belong to?

What clusters do the above instance groups belong to and how are we determining that its ok to GC them?

For example, what was the purpose of the GKE cluster named "gke-apps-apps-cpu-pool-v1-a3e58689-grp" ? How do we know its safe to GC it?

The IG named "gke-kf-vmaster-1124--kf-vmaster-1124--fae5daa0-grp" belongs to an auto-deployed cluster; so we should be applying the criterion for the auto-deployed clusters.

PR #529, opened 2019-11-25
Gc instance groups based on usage

gabrielwen:
Fixes: #528 
Verified this should work locally.;

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/529)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -304,9 +304,6 @@ def cleanup_instance_groups(args):
   deleted = []
   in_use = []
 
-  # TODO(jlewi): We should check whether the instance group is in use
-  # before deleting it. At least in pantheon it looks like instance groups
-  # are listed as in use by clusters.
   for zone in args.zones.split(","): # pylint: disable=too-many-nested-blocks
     while True:
       results = instanceGroups.list(project=args.project,
@@ -316,7 +313,6 @@ def cleanup_instance_groups(args):
         break
       for s in results["items"]:
         name = s["name"]
-        age = getAge(s["creationTimestamp"])
         size = s["size"]
         if size > 0:
           logging.info("Skipping instance group %s because it is in use by %d "
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -20,11 +20,11 @@ from oauth2client.client import GoogleCredentials
 # See https://github.com/kubeflow/testing/issues/444
 # We are switching to unique names for auto deployments
 # So this matches the new ones.
-AUTO_DEPLOY_PATTERNS = [re.compile(r"kf-vmaster-(?!n\d\d)")]
+AUTO_DEPLOY_PATTERNS = [re.compile(r".*kf-vmaster-(?!n\d\d)")]
 
 E2E_PATTERNS = [re.compile(".*e2e-.*"), re.compile(".*kfctl.*"),
                 re.compile(".*z-.*"), re.compile(".*presubmit.*"),
-                re.compile(".*unittest.*")]
+                re.compile(".*unittest.*"), re.compile("k8s-ig-.*")]
 
 # Constants enumerating the different classes of infra
 # We currently have 2 types
@@ -305,16 +305,6 @@ def cleanup_instance_groups(args):
   unexpired = []
   in_use = []
 
-  auto_deploy_patterns = re.compile("gke-kf-vmaster.*")
-  e2e_pattern = re.compile("k8s-ig-.*")
-
-  def ig_name_to_infra_type(name):
-    if auto_deploy_patterns.match(name):
-      return AUTO_INFRA
-    if e2e_pattern.match(name):
-      return E2E_INFRA
-    return E2E_OWNERLESS
-
   for zone in args.zones.split(","): # pylint: disable=too-many-nested-blocks
     while True:
       results = instanceGroups.list(project=args.project,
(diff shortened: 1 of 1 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -316,39 +313,30 @@ def cleanup_instance_groups(args):
...
+        size = s["size"]
+        if size > 0:
+          logging.info("Skipping instance group %s because it is in use by %d "
+                       "instances.", name, size)
+          in_use.append(name)
           continue
 
-        logging.info("Instance group %s categorized as %s", name, infra_type)
-
-        if age > MAX_LIFETIME[infra_type]:
```
jlewi: Why are you removing the max lifetime criterion? What if a GKE cluster orphans an instance group of fixed size how would that end up being GC'd?

jlewi:
@gabrielwen Could you please update the PR description to explain the issue and how this fixes it? In particular why are we current leaking clusters and how is this new policy going to delete IG's we want to GC but not those we don't?

In #528 one of the listed IGs "gke-kf-vmaster-1124--kf-vmaster-1124--fae5daa0-grp" should be long to  an auto-deployed cluster; so we should be applying the policy around auto-deployed clusters to that IG. How come that cluster didn't match?

gabrielwen:
> @gabrielwen Could you please update the PR description to explain the issue and how this fixes it? In particular why are we current leaking clusters and how is this new policy going to delete IG's we want to GC but not those we don't?
> 
> In #528 one of the listed IGs "gke-kf-vmaster-1124--kf-vmaster-1124--fae5daa0-grp" should be long to an auto-deployed cluster; so we should be applying the policy around auto-deployed clusters to that IG. How come that cluster didn't match?

Because it has `gke-` as prefix so the regex couldn't identify it.

jlewi:
@gabrielwen Should we adjust the regex?

py/kubeflow/testing/cleanup_ci.py
gabrielwen: move the policy back.  to the question, I think as long as an instance group has dependencies running, we cannot delete it.

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -293,20 +293,28 @@ def cleanup_firewall_rules(args):
...
   next_page_token = None
-  expired = []
+  deleted = []
   unexpired = []
   in_use = []
 
-  # TODO(jlewi): We should check whether the instance group is in use
-  # before deleting it. At least in pantheon it looks like instance groups
-  # are listed as in use by clusters.
+  auto_deploy_patterns = re.compile("gke-kf-vmaster.*")
```
jlewi: Why not just change the existing regex from "kf-vmaster.*" to ".*kf-vmaster.*"?

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -293,20 +293,28 @@ def cleanup_firewall_rules(args):
...
+  e2e_pattern = re.compile("k8s-ig-.*")
+
+  def ig_name_to_infra_type(name):
```
jlewi: Why not just modify name_to_infra to return E2E_OWNERLESS if the name matches various regex and then define OWNERLESS_PATTERNS that includes "k8s-ig.*"

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -317,38 +325,38 @@ def cleanup_instance_groups(args):
...
-                                               zone=zone,
-                                              instanceGroup=name).execute()
-              logging.info("response = %r", response)
-              expired.append(name)
-            except Exception as e: # pylint: disable=broad-except
-              logging.error(e)
-              in_use.append(name)
-        else:
+        infra_type = ig_name_to_infra_type(name)
+        logging.info("Instance group %s has been identified as %s", name, infra_type)
```
jlewi: It looks like you eliminated the check that infra_type is not none. Should we still be enforcing that? i.e. we only delete instances if we are able to map them to an infra type?

gabrielwen: done.

py/kubeflow/testing/cleanup_ci.py
gabrielwen: changed.

py/kubeflow/testing/cleanup_ci.py
gabrielwen: updated.

py/kubeflow/testing/cleanup_ci.py
```diff
@@ -317,38 +315,41 @@ def cleanup_instance_groups(args):
       for s in results["items"]:
         name = s["name"]
         age = getAge(s["creationTimestamp"])
+        size = s["size"]
+        if size > 0:
```
jlewi: Doesn't this make our script less restrictive?
Right now we GC IG's after they expire. You are now skipping IG's which have expired but have non-zero size. 

If IUC there were two sets of IG's that we were leaking before because the regex didn't match.
k8s-ig.*
gke-kf-vmaster.*

You've updated the regex to match these IGs  So that should take care of the leak.

So separate from that do you also want to guard against GC'ing IGs with size > 0 ?

gabrielwen: I think if the size > 0, we are not able to delete the IG anyway because of resource dependency.

To the scenario you described, cleanup_ci will first delete the cluster, making the size to 0, IG will be cleaned up during the same run or next.

jlewi: @gabrielwen Thanks. Looks like we do set the min pool size to zero
https://github.com/kubeflow/manifests/blob/00fa81b956682ee835d75e21413c8008bf2bf170/gcp/deployment_manager_configs/cluster-kubeflow.yaml#L53

Issue #450, opened 2019-08-29
Continuous build of docker images and updating kustomize manifests


jlewi:
We need a good way to continuously build our docker images and then update our kustomize manifests to use the updated images.

This is critical for maintaining velocity. One of the big problems we are seeing with releases is that changes are piling up and not getting exercised until we start cutting releases because we haven't updated our kustomize manifests.

Also as the number of applications scale the toil around building docker images and then updating manifests becomes significant. This is especially true during releases as we try to rapidly push out fixes.

There is a POC based on the jupyter web app here.
https://github.com/kubeflow/kubeflow/tree/master/releasing/auto-update

We'd like to make it super easy for people to define new workflows to auto-build their application. In an ideal world they would just check in a YAMl file with a couple configurations e.g.

* Location of their Dockerfile
* Location of their kustomization.yaml file

A couple things we are missing

1. A good solution for triggering workflows on pre/postsubmits/cron jobs
1. A good solution for monitoring/alerting
1. A good story for reusability around common tasks (e.g. building images, creating a PR, etc...)

/cc @scottilee @animeshsingh @kkasravi @jinchihe

jlewi:
GCB now has direct integration via GitHub App Triggers
https://cloud.google.com/cloud-build/docs/create-github-app-triggers

So if we install that GitHub App in our project then we can trigger GCB builds in response to PRs.  The GCB build could then create K8s resources. 

This is very similar to our prow infra works today. We use Prow to trigger Prow jobs which run run_workflow_e2e.py which in turn submits a bunch of Argo workflows based on prow_config.yaml.

We could do something similar but use GCB to invoke run_e2e_workflow.py

jlewi:
With kubeflow/kubeflow#4029 we have a pretty good POC for CD of the jupyter web app image.
The next step is probably to generalize this to a 2nd image which should probably be the central dashboard; kubeflow/kubeflow#3781.

1. Parameterizing [update_jupyter_web_app.py](https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/update_jupyter_web_app.py) into a reusable script should be pretty easy

    * It looks like there are basically a couple arguments
      1. The image location
      1.  Kustomization location
      1.  Location of source and the command to build the image
           * We could make certain assumptions; i.e. that there is a Makefile and the name of the build rule and variables parameterizing where it gets pushed

1. Make it easy for people to add new jobs to periodically build and push their image.

[...]

PR #561, opened 2020-01-09
update_kf_apps.py should create pipelineruns for images that need to be updated.

jlewi:
update_kf_apps.py should create pipelineruns for images that need to be updated.

* Determine whether an image is already up to date by comparing the desired
  image to the image listed in the manifest

* If the image needs to be updated submit the PipelineRun to update the
  image.

* Related to #450

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/561)

### DIFF
--- a/py/kubeflow/testing/cd/update_kf_apps.py
+++ b/py/kubeflow/testing/cd/update_kf_apps.py
@@ -361,8 +361,6 @@ class UpdateKfApps:
       logging.info("No pipelines need to be run")
       return
 
-    #pipelines_to_run = ["/tmp/runs/admission-webhook-run-master-6b987df8.yaml",
-                        #"/tmp/runs/kfam-run-master-caf5eb33.yaml"]
     for p in pipelines_to_run:
       with open(p) as hf:
         run = yaml.load(hf)
--- a/py/kubeflow/testing/cd/update_kf_apps.py
+++ b/py/kubeflow/testing/cd/update_kf_apps.py
@@ -8,12 +8,19 @@ import re
 import yaml
 
 from kubeflow.testing import util
+from kubernetes import client as k8s_client
+from kubernetes import config as k8s_config
+from kubernetes.client import rest
 
 GIT_URL_RE = re.compile(r"([^:]*):([^/]*)/([^\.]*)\.git")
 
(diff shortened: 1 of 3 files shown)


jlewi:
/assign @zhenghuiwang 

This is ready for review.

py/kubeflow/testing/cd/update_kf_apps.py
```diff
@@ -229,6 +338,77 @@ def create_runs(config, output_dir, template, src_dir):
...
+    crd_api = k8s_client.CustomObjectsApi(client)
+
+    _, pipelines_to_run = UpdateKfApps.create_runs(
+      config, output_dir, template, src_dir)
+
+    if not pipelines_to_run:
+      logging.info("No pipelines need to be run")
+      return
+
+    #pipelines_to_run = ["/tmp/runs/admission-webhook-run-master-6b987df8.yaml",
```
zhenghuiwang: Delete this line?

py/kubeflow/testing/cd/update_kf_apps.py
```diff
@@ -229,6 +338,77 @@ def create_runs(config, output_dir, template, src_dir):
...
+                        #"/tmp/runs/kfam-run-master-caf5eb33.yaml"]
+    for p in pipelines_to_run:
+      with open(p) as hf:
+        run = yaml.load(hf)
+
+      group, version = run["apiVersion"].split("/", 1)
+      kind = run["kind"]
+      plural = kind.lower() + "s"
```
zhenghuiwang: I'm wondering if this is a robust way to get kind plural, although it should work in most cases.

jlewi: Since its a custom resource in this case we could do `kubectl get crd` and the plurals is defined in the CRD. At this point though its not really necessary since we really only work with tekton pipeline runs which we know the heuristic works for.

I'm not sure what you would do for built in objects; e.g. Deployment since there is no CRD.

zhenghuiwang: ah, I missed they are tekton runs but was thinking of all possible kinds.

PR #566, opened 2020-01-21
Fix cleanup logic for IAM policy bindings

jlewi:
* We are not properly GC'ing policy bindings for deleted service accounts.

* The problem is that we only consider service accounts matching a certain
  regex and that regex isn't matching service accounts for our auto-deployed
  clusters.

* Using a regex should be unnecessary. If a service account doesn't exist
  that should be a sufficient criterion that the policy bindings should be
  deleted.

Related to: kubeflow/testing#543

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/566)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -759,18 +759,23 @@ def cleanup_service_accounts(args):
 
 def trim_unused_bindings(iamPolicy, accounts):
   keepBindings = []
+  deleted_bindings = set()
+  kept_bindings = set()
   for binding in iamPolicy['bindings']:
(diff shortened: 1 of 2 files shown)


jlewi:
Here are the logs from a one off run. 
[oneoff-run.txt](https://github.com/kubeflow/testing/files/4093600/oneoff-run.txt)

It looks like a bunch of old bindings were GC'd.

Here's the current policy.

[kubeflow.ci.deployment.policy.txt](https://github.com/kubeflow/testing/files/4093608/kubeflow.ci.deployment.policy.txt)

Looks like a bunch of bindings were pruned.

PR #568, opened 2020-01-21
Revert "Fix cleanup logic for IAM policy bindings (#566)"

jlewi:
This reverts commit 4ebe57134446afe6aa5bbd33494d91831035a30f.

The list of service accounts only includes service accounts owned by the
project. So we ended up deleting bindings for service accounts owned by
other projects such as the service accounts.

related to: #566

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/568)


jlewi:
Force merging because tests are broken right now because of this.

PR #570, opened 2020-01-22
Fix cleanup logic for IAM policy bindings.

jlewi:
* We want to cleanup bindings for service accounts that don't exist
  as a way to GC any bindings that might otherwise accumulate.

* However, we only want to GC bindings for service accounts that are
  owned by that project because only those service accounts will be listed
  as service accounts for that project

  e.g. in the project kubeflow-ci-deployment; we should only delete bindings
  for the service accounts
  ${NAME}@kubeflow-ci-deployment.iam.gserviceaccount.com

  Any other bindings should be preserved.

* related to: #566

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/570)

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -1,6 +1,7 @@
 """Cleanup Kubeflow deployments in our ci system."""
 # pylint: disable=too-many-lines
 import argparse
+import collections
 import datetime
 from dateutil import parser as date_parser
 import logging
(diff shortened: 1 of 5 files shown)


jlewi:
Here are drydrun logs

[cleanup.dryrun.logs.txt](https://github.com/kubeflow/testing/files/4094861/cleanup.dryrun.logs.txt)

The logs show that we don't cleanup bindings for service accounts in other projects e.g. bindings for `kubeflow-testing@kubeflow-ci.iam.gserviceaccount.com` are preserved.

jlewi:
pylint failed

Issue #571, opened 2020-01-22
CD scripts for updating kubeflow manifests should close older PRs


jlewi:
Follow on to #450 

We have a server which continually opens up PRs to update the Kubeflow manifests to a newly built docker image for each application
https://github.com/kubeflow/testing/tree/master/apps-cd

Its not uncommon for us to end up with multiple PRs in flight for a given application because there were multiple commits to an application.

In this case, we should automatically close all but the most recent PR for an application so as not to confuse the reviewer.

jlewi:
It looks like the hub CLI doesn't support closing PRs so we will need to use the GitHub API to do this.

I think we will want to create a python script which uses a GitHub App to comment and close PRs.

Issue #100, opened 2018-04-05
[MAINTENANCE] Expand kubeflow-ci cluster to add gpus and cpus.


elsonrodriguez:
This'll cause a momentary blip.

elsonrodriguez:
This is done.

Old Cluster size:
3X n1-standard-8
1x  n1-standard8,2xK80

New cluster size:
4X n1-standard-8
2x  n1-standard8,2xK80

PR #575, opened 2020-01-24
Cicd delete old prs

jlewi:
Define a script to close obsolete PRs to update an application.

* In the event there are multiple PRs open to update a Kubeflow application
  we want to close the older PRs; so there is a single open PR updating
  the application to the newest code.

  Related to #571

* Setup a def namespace for use with apps-cd.

* Update update_kf_apps.py to close old PRs on each sync.

* Bake the source code into the docker image rather than using a wrapper
  script to sync the code from git.

  * Sync'ing the code from git became to difficult to reason about once
    we start splitting the source code across multiple repositories
    * We now depend on github/kubeflow/code-intelligence for utilities
      for working with GitHub Apps

    * Using a docker image also ensures we don't get broken suddenly when
      new changes are in place

    * In the future we could use github actions to automate updating the
      deployment on postsubmits

* Turn app-pipeline.template.yaml into a ConfigMap
  * This allows better versioning
  * We can rely on kustomize to create a configmap with a hash based on the
    contents
  * kustomize will then reference the config map using its hash. As
    a result a rolling update is triggered whenever the hash contents changes.
  * This makes it easier to handle rollous and updates.

Define a dev instance of the update KF apps infrastructure to facilitate development
  * Use profiles in skaffold.

  * update_kf_apps.py in dev uses a config map now to ubtain app-pipeline.template.yaml 
    rather than fetching it from git

  * This makes it much easier to test changes in the dev instance

Fix a bunch of bugs preventing update_kf_apps.py from working 
   * Update requirements.txt with a bunch of missing packages.
   * Fix some imports in update_kf_apps.py

  * Need to set resource requests for the build pods otherwise builds get
    CPU starved and take forever.

Miscellaneous

* Create a tool to copy secrets between namespaces from GCS

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/575)


py/kubeflow/testing/cd/close_old_prs.py
```diff
@@ -0,0 +1,225 @@
...
+        app_prs[app] = app_prs[app] + [pr]
+
+    # For each application and tuple close all but the most recent PRs
+    for app, prs in app_prs.items():
+      logging.info(f"{app} has {len(prs)} open prs")
+      if len(prs) == 1:
+        continue
+
+      # We sort the PRs by URL since this will correspond to PR number
+      # We will keep the most recent one.
```
zhenghuiwang: What does the PR url looks like?

If URLs are in the format
`https://github.com/kubeflow/testing/pull/99`
`https://github.com/kubeflow/testing/pull/100`

Sorting lexicographically yields unexpected outcome?

jlewi: Good catch. I'll fix this.

jlewi:
/hold
I'm refactoring this to better handle how we delete old PRs.

jlewi:
I've rebased the changes and updated the PR description.

Here's an example PR created by the dev instance
kubeflow/manifests/pull/754

Unfortunately the tests are failing due to #582

PR #587, opened 2020-01-29
Add TFJob and PyTorch controller to the list of applications for CD.

jlewi:
* Update the Tekton pipelines to support applications whose source
  code isn't in the kubeflow repository (e.g. TFJob controller and pytorch)

  * We need to generalize the name of the git resource containing
    the application to be app-repo.

* update_kf_apps.py needs to filter out repositories that aren't need for
  a particular application; attaching extra resources to a pipeline causes
  the run to fail.

* Update tests to work with assumption that source could be a different
  repository.

* Use structured logging to make it easier to find logs about a particular
  application

Related to #450

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/587)

### DIFF
--- a/apps-cd/applications.yaml
+++ b/apps-cd/applications.yaml
@@ -73,6 +73,30 @@ applications:
       value: "gcr.io/kubeflow-images-public/profile-controller"
     # The name of the repo containing the source
     sourceRepo: kubeflow
+  - name: pytorch-operator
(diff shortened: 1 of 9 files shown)


jlewi:
Here's a sample PR.
https://github.com/kubeflow/manifests/pull/787

PR #596, opened 2020-02-04
Create a script to make updating application resources easier.

jlewi:
* This is a replacement for
  https://github.com/kubeflow/manifests/blob/master/hack/update-instance-labels.sh

* We don't want to assume all applications are at the same version
* This script makes it easier to set a different version for a specific
  application.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/596)

### DIFF
--- a/py/kubeflow/testing/tools/applications.py
+++ b/py/kubeflow/testing/tools/applications.py
@@ -1,0 +1,79 @@
+"""A CLI for updating YAML files defining application CRs.
+
+This tool makes it easier to update and maintain the Application CRs
+for each application.
+"""
+
+import fire
+import logging
+import os
+import yaml
+
+class AppManager:
+  @staticmethod
+  def update(version, directory=None):
+    """Update the application resources to the specified version.
+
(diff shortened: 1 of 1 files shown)


zhenghuiwang:
/lgtm

I'm wondering if this script is more discoverable if we put it into kubeflow/manifests/hack and add a Makefile target for it.

jlewi:
@zhenghuiwang that's a good question.

My current thinking is that we should try to concentrate all of our release/ci tooling in kubeflow/testing rather than spreading it out in multiple repositories.

Putting all the python code in one repository makes it easier to reuse common libraries.

zhenghuiwang:
> @zhenghuiwang that's a good question.
> 
> My current thinking is that we should try to concentrate all of our release/ci tooling in kubeflow/testing rather than spreading it out in multiple repositories.
> 
> Putting all the python code in one repository makes it easier to reuse common libraries.

Putting everything into a single place makes sense.

/lgtm

PR #597, opened 2020-02-04
update default pattern to match new convention

gabrielwen:
seems like the new convention is `kf-master` instead of `kf-vmaster`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/597)


jlewi:
This works.
Note we also set labels on the deployments and those labels provide a way to match clusters. in particular the label "git-manifests-branch" will be set to the name of the branch from which we deployed Kubeflow.

Ultimately that might be a better way to select auto deployed clusters.

Issue #595, opened 2020-02-03
Auto-deploy V2


jlewi:
This is the uber bug for updating the auto-deployment infrastructure.

The key feature that we'd like to achieve with V2 is to auto-deploy as soon as a PR is merged.

The current implementation doesn't facilitate that because we have a simple cron job which launches a new deployment every X hours.

If we decrease X in order to deploy sooner we will end up accumulating multiple instances of the auto-deployed infra and using up all our quota.

We want more of a rate limiting queue type pattern similar to what K8s controllers and reconcilers do.

This is critical to being able to rapidly verify changes to the deployments.

jlewi:
I created a new cluster on which to deploy this

*  **project**: kubeflow-ci
*  **cluster**: kubeflow-ci-v1

This is a V1 Kubeflow cluster.

I couldn't use the existing kubeflow-testing cluster because it doesn't have ISTIO installed. I want to use ISTIO to be able to setup routes to the web app we will create.

jlewi:
I think the cleanup script might have deleted some instances.

The endpoint when down and now its back up but I'm getting origin authentication failures.

jlewi:
I deleted the iap-enabler pod and that fixed it. 
It looks like the iap-enabler pod got stuck trying to get the backend name
https://github.com/kubeflow/manifests/blob/f239b294879f9523248106ebb74f6c12e5d33765/gcp/iap-ingress/base/config-map.yaml#L85

The backend name kept being empty.

jlewi:
I deleted the old cron job `auto-deploy-master`.

jlewi:
I deleted the cron job to cleanup kubeflow-ci

```
kubectl get cronjobs
NAME                                SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
cleanup-ci                          0 */2 * * *   False     0        71d             323d
cleanup-ci-kubeflow-ci              0 */2 * * *   False     0        66m             321d
cleanup-ci-kubeflow-ci-deployment   0 */2 * * *   False     0        66m             321d
jlewi@jlewi-glaptop2:~/git_kubeflow-testing$ kubectl delete cronjobs cleanup-ci-kubeflow-ci
cronjob.batch "cleanup-ci-kubeflow-ci" deleted
jlewi@jlewi-glaptop2:~/git_kubeflow-testing$ kubectl delete cronjobs cleanup-ci
cronjob.batch "cleanup-ci" deleted
```

[...]

PR #603, opened 2020-02-10
Create a new Kubeflow cluster to run auto-deploy v2.

jlewi:
* #595 we are creating a new version of the infrastructure to auto-deploy
  Kubeflow from master and release branches

* This PR contains the GCP and Kustomize manifests for setting up a new
  Kubeflow cluster where this app will run.

* We can't use the existing kubeflow-testing cluster because that doesn't
  have ISTIO or workload identity which we want for the auto-deploy app.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/603)


jlewi:
@zhenghuiwang I removed some OWNERs files; could you please re-LGTM.

Issue #560, opened 2020-01-08
Update pylint in the test worker image not compatible with f style strings


jlewi:
I get different behavior when running pylint in the test worker image then when running it locally.

Example PR
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_testing/559/kubeflow-testing-presubmit/1214777447178833920/

When run in the worker image I get

```
util.py                     72 INFO     ************* Module kubeflow.testing.cd.update_kf_apps
util.py                     72 INFO     E: 45, 0: invalid syntax (<string>, line 45) (syntax-error)
util.py                     46 INFO     Running: pylint --rcfile=/mnt/test-data-volume/kubeflow-testing-presubmit-py-unittests-559-d51aded-3920-83d3/src/kubeflow/testing/.pylintrc /mnt/test-data-volume/kubeflow-testing-presubmit-py-unittests-559-d51ade
d-3920-83d3/src/kubeflow/testing/py/kubeflow/testing/cd/update_kf_apps_test.py
```

I do not get that error when running locally.

Locally I'm using version

```
pylint3 --version
pylint3 2.2.2
astroid 2.1.0
Python 3.7.5rc1 (default, Oct  2 2019, 04:19:31) 
[GCC 8.3.0]
```
in gcr.io/kubeflow-ci/test-worker@sha256:fbfb9141c02e9a50e3679f84637fc2d7c142fe9eaa6b9f55a3c9f8a710000837
```
pylint --version
No config file found, using default configuration
pylint 1.8.2, 
astroid 1.6.1
Python 3.5.2 (default, Oct  8 2019, 13:06:37) 
[GCC 5.4.0 20160609]
```

I tried installing pylint3 in the container and got the same error

```
pylint3 --version
Using config file /git_kubeflow-testing/.pylintrc
pylint3 1.8.2, 
astroid 1.6.1
Python 3.5.2 (default, Oct  8 2019, 13:06:37) 
[GCC 5.4.0 20160609]
```

It looks like the syntax error is caused by f-style strings.

e.g.

```
  raise ValueError(f"Repository {r['name']} is missing param url") # pylint: disable=syntax-error
```

I suspect to fix this we may need to upgrade the python version to 3.7

jlewi:
I think we can do something like the following to install the later version of python3

```
RUN apt-get update -y && \
    apt-get install -y python3.8 python3-pip && \
    ln -sf /usr/bin/python3.8 /usr/bin/python

[...]

PR #606, opened 2020-02-11
Create a python3.8 version of the worker image

jlewi:
* Related to kubeflow/testing#560 - python3 lint
* We need this because we need a newer (3.8) version of pylint
  to allow for python3.8 syntax like f strings.

* In this pass we eschewed the following packages which I don't think we
  need anymore

  * ksonnet
  * helm
  * glide (should be using go modules and staged builds)
  * docker
  * nodejs (should be using staged builds)

* Create a skaffold config to build the image.

* Related to: kubeflow/kfctl#228 endpoint test is failing

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/606)

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,5 +1,7 @@
 # Docker image for running E2E tests using Argo.
-
+#
+# This image is obsolete
+# we should start using Dockerfile.py3 which has python3
 FROM ubuntu:xenial
 MAINTAINER Jeremy Lewi
 
--- a/images/Dockerfile.py3
+++ b/images/Dockerfile.py3
@@ -1,0 +1,64 @@
+# Build the docker image used to run the scripts
+# to continuously update our docker files.
+#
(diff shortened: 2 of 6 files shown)


jlewi:
Running pylint in the docker image

```
pylint --version
pylint 2.4.4
astroid 2.3.3
Python 3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 8.3.0]
```

zhenghuiwang:
Using Python3.8 and Ubuntu 18.04 is great! 

Do we need to apt-get these packages (build-essential \cmake \ curl etc) as before?

jlewi:
@gabrielwen added gcloud

@zhenghuiwang curl is already there; I don't think we use cmake anywhere. I'm not sure if we need build-essential. If we are building binaries  We use make and that's already installed. What's in build-essential?

jlewi:
@gabrielwen any additional comments?

gabrielwen:
no
/lgtm

PR #615, opened 2020-02-19
Fix py3 test worker

gabrielwen:
Adding additional installations for the test worker to work on tests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/615)

### DIFF
--- a/images/Dockerfile.py3
+++ b/images/Dockerfile.py3
@@ -14,9 +14,10 @@ RUN python3.8 -m pip install \
     watchdog \
     pytest \
     kubernetes \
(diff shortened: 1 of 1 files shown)


images/Dockerfile.py3
```diff
@@ -11,7 +11,18 @@ RUN apt-get update -y && \
 RUN python3.8 -m pip install \
     fire \
     lint \
-    watchdog
+    watchdog \
```
jlewi: nit alphabetize

gabrielwen: done.

PR #622, opened 2020-03-05
Tekton prow scripts

gabrielwen:
Implements Tekton runner to run Tekton pipelines.  Also provides examples of Task in templates and how to link from Prow to the tests in prow_config.yaml and PipelineRun configs.

Related to #425 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/622)


gabrielwen:
progress as of today: steps in a task will stop running if any of previous steps has failed.  this is a WIP feature in Tekton to solve.  So we still need to use PVC to pass artifacts to another task for copy to jenkins GCS.  hopefully this will cause less problems as the files are small.  still, within a DAG if previous task fails, the following task doesn't seem to be running as well.  need to find a way to enforce it.

gabrielwen:
@jlewi this is ready.

gabrielwen:
@jlewi not yet - failed test junit xml is not able to upload to GCS, looking into it.

gabrielwen:
@jlewi ok it's ready for review.

gabrielwen:
hold on, I just found out I need to update the workflow to work with post-submits and periodic tests.

jlewi:
Really great to see this coming along.

Could you call out in the PR description some of the limitations; e.g.
i) the need for PVC to work around that a failed step will block subsequent steps and tasks
   * Can we eliminate that by using Tekton's GCS resource to upload the artifacts from /workspace?

ii) The work around for teardown steps

py/kubeflow/testing/run_e2e_workflow.py
```diff
@@ -343,6 +350,86 @@ def run(args, file_handler): # pylint: disable=too-many-statements,too-many-bran
...
+        test_target_name = workflow_name
+      logging.info("Reading Tekton PipelineRun config: %s", test_target_name)
+      tekton_run["metadata"]["name"] = workflow_name
+
+      for t in tekton_run.get("spec", {}).get("pipelineSpec", {}).get("tasks", []):
+        if not "params" in t:
+          t["params"] = []
+
+        # Update Junit artifacts XML.
+        for i in range(len(t["params"])):
```
jlewi: Could you explain why run_e2e_workflow.py has to adjust the path of the junit files?

tekton/configs/junit-pvc.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
jlewi: Is it 1 PVC per task? per pipeline run?

tekton/templates/debug-task.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: tekton.dev/v1alpha1
```
jlewi: Could you add a comment explaining what this task is for?

tekton/templates/notebook-test-task.yaml
```diff
@@ -8,6 +6,8 @@ metadata:
   annotations:
     sidecar.istio.io/inject: "false"
 spec:
+  workspaces:
+  - name: junit-artifacts
```
jlewi: What are Tekton workspaces and how are they being used?

tekton/templates/test-teardown-task.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - name: build-id
+      type: string
+      description:
+        BUILD_ID from Prow ENV.
+    resources:
+    - name: kf-testing-repo
+      type: git
+      targetPath: src/kubeflow/testing
+  steps:
+  - name: copy-artifacts
```
jlewi: Can we use Tekton resources to upload files to GCS as opposed to having a separate task for that?

jlewi:
@gabrielwen let me know when this is ready for another look.

tekton/templates/test-teardown-task.yaml
gabrielwen: just confirmed we could.

gabrielwen:
@jlewi I think I got what we need for the design.  let's discuss in the doc and update here later on when we have a concrete version.  will share the doc in 1 - 2 days.

jlewi:
@gabrielwen What is the status of this? When do you think it will be ready for review?

gabrielwen:
@jlewi this is ready.  however, it seems all the tests are failing.  could you take a look and force merge it when it's ready?

jlewi:
@gabrielwen Could you update the PR description to explain what this PR is doing? Use the PR description to help the reviewer figure out what they are looking at and what they should focus on.

jlewi:
This is really great. Excited to get this in.

jlewi:
@gabrielwen Any chance when another round will be ready for review?

jlewi:
/lgtm
/approve
I think we can merge this as is and address the rest of the comments in follow on PRs. 

It looks like the Tekton test for mnist is failing. We probably shouldn't be running that test for kubeflow/testing anyway. A lint or unittest would probably make more sense. Maybe we can just remove those tests.

jlewi:
/lgtm
/approve

Tests will probably fail because of #659 once that is fixed we should be able to merge this.

PR #623, opened 2020-03-14
Added gsoc label to kubeflow_label.yml

sarahmaddox:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/623)


jlewi:
@sarahmaddox I would suggest changing the label to "area/gsoc"

The main reason is that this will allow the label to be added via chat-bot commands as you can see from my comment above. This works for area labels but not community labels.

sarahmaddox:
Thanks @jlewi PTAL.

Issue #624, opened 2020-03-17
label sync needs to auto-sync the config to the cluster


jlewi:
Our label sync process currently relies on manually updating the config map when things change.
https://github.com/kubeflow/testing/tree/master/label_sync

Lets try to use gitops and anthos CM for this.
https://cloud.google.com/anthos-config-management/docs/quickstart

Swikar:
Jeremy, I noticed Anthos setup is done. I am not sure fully completed or not? Can you tell me what portion of work still require for this PR?

Swikar:
@jlewi Regarding https://github.com/kubeflow/testing/issues/624 How far has been completed and how much work still reamains? I am currently working on that PR. I am reviewing this https://cloud.google.com/anthos-config-management/docs/quickstart (

jlewi:
@Swikar I think this was fixed by #625

Swikar:
Oh okay it seemed to me👍

PR #625, opened 2020-03-18
Use Anthos Config Management to continually sync the label config to the cluster

jlewi:
* Define a kustomize package to generate the config map from the labels

* ACM requires the directory structure follow a certain format so we
  created the directory label_sync/acm_repo

* A simple shell script runs kustomize to dump the manifests to the directory.

Related to #624

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/testing/625)

### DIFF
--- a/label_sync/acm_repo/namespaces/batch_v1beta1_cronjob_label-sync-cron.yaml
+++ b/label_sync/acm_repo/namespaces/batch_v1beta1_cronjob_label-sync-cron.yaml
@@ -1,0 +1,37 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: label-sync-cron
+spec:
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    metadata:
+      labels:
+        app: label-sync
+    spec:
+      template:
+        spec:
+          containers:
+          - args:
+            - --config=/etc/config/kubeflow_label.yml
+            - --confirm=true
+            - --orgs=kubeflow
+            - --token=/etc/github/bot-token
+            image: gcr.io/k8s-testimages/label_sync:v20180921-f7ff24f34
+            name: label-sync
+            volumeMounts:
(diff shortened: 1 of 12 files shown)


label_sync/label_sync_acm_config.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  # name: config-management
+  name: label-sync
+  # Does this need to be in the same namespace as the configmanagement system?
+  namespace: config-management-system
+spec:
+  # clusterName is required and must be unique among all managed clusters
+  clusterName: my-cluster
+  git:
+    # TODO(jlewi): Change to master before committing
```
gabrielwen: it's done?

label_sync/acm_repo/namespaces/batch_v1beta1_cronjob_label-sync-cron.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: batch/v1beta1
```
gabrielwen: for my own education: what is the difference between this cron and cluster/label_sync_cron_job.yaml?

jlewi: The files in cluster is a kustomization package which is used to generate the "hydrated" manifests (kustomize build -o output).

We use kustomize because for the configmap we want to generate the configmap from an external file that contains the contents of the configmap but not the configmap itself.

Furthermore, we want the configmap to have a unique name based on the content hash of the configmap. This way any resource (e.g. a deployment) which references the configmap would change when the content of the config map changes which will cause new pods to be deployed. If we don't do this (i.e. if the configmap name is not unique based on the hash) then pods would not necessarily pick up the changed configmap.

By extension we use kustomize to generate the cron job because kustomize will correctly substitute in the generated name of the configmap which includes the unique content hash.

Issue #636, opened 2020-04-10
Auto-deployed instances from master and v1 branch are not accessible; out of ssl certificate quota


jlewi:
Tried the endpoints:
https://kf-master-0410-b46.endpoints.kubeflow-ci-deployment.cloud.goog/
https://kf-v1-0410-656.endpoints.kubeflow-ci-deployment.cloud.goog/

Get ERR_CONNECTION_CLOSED

jlewi:
The latest endpoints are working.

https://kf-master-0413-353.endpoints.kubeflow-ci-deployment.cloud.goog/

I suspect we hit SSL quota issues. It doesn't look like they are using self-signed ceritificates.

jlewi:
Endpoint:
https://kf-master-0413-6eb.endpoints.kubeflow-ci-deployment.cloud.goog/

Here's the ingress resource
[ingress-kf-master-0413-6eb.yaml.txt](https://github.com/kubeflow/testing/files/4469889/ingress-kf-master-0413-6eb.yaml.txt)

   * Looks like its still using a GKE managed certificate

The deploy logs look like it is creating a self signed certificate
[auto-deploy-master-mtwg6-rx9lp.log.txt](https://github.com/kubeflow/testing/files/4469903/auto-deploy-master-mtwg6-rx9lp.log.txt)

https://github.com/kubeflow/testing/blob/52b4f0b6539df31582a84f1b17b9b4acc6462f0d/py/kubeflow/testing/util.py#L825

jlewi:
The annotations:

* ingress.gcp.kubernetes.io/pre-shared-cert
* networking.gke.io/managed-certificate

Are still there.

jlewi:
Remove the annotation

```
kubectl -n istio-system annotate ingress envoy-ingress --overwrite=true "networking.gke.io/managed-certificates"

kubectl -n istio-system annotate ingress envoy-ingress --overwrite=true ingress.gcp.kubernetes.io/pre-shared-cert- 
```
  * I think we need to remove "network.gke.io/..." first otherwise the "ingress.gcp.kubernetes.io/pre-shared-cert-" will just get added back automatically

jlewi:
After deleting those annotations the self-signed certificate got used.

jlewi:
Lets Encrypt quota should no longer be an issue. Since its working now its unclear what the actual problem was. So I'm going to close this as not reproducible for now.

I'm also going to go ahead and update #637 to remove the use of self signed ceritifcates.

[...]

PR #637, opened 2020-04-13
Don't use self signed certificates in auto deployed infrastructure

jlewi:
* We should no longer need to use self signed ceritificates with auto-deployed instances because
  GKE ManagedCertificates don't have the SSL quota limits of lets encrypt.

* Nonethless fix a bug in the logic for attaching a self signed ceritificate.
   * We need to delete the managed certificate annotations otherwise
      the ingress doesn't end up using the self signed ceritifcate and
      we end up running out of SSL quota.

* Creating a playbook directory and start putting the playbooks there.

* Related to kubeflow/testing#636


jlewi:
/hold

It looks like managed certificates should have higher quotas now.

jlewi:
@kunmingg I think this is ready for review.

PR #646, opened 2020-04-29
Remove applications.py; commonLabels should be immutable.

jlewi:
* See kubeflow/manifests#1131  we want commonLabels to be immutable
  otherwise upgrades won't work because we end up trying to change
  selectors which are immutable.

* The applications.py script inherently violates that because it was
  mutating commonLabels on every release. We don't want to do that
  anymore so we delete the script.


yanniszark:
@jlewi thanks for the PR!
/lgtm

PR #653, opened 2020-05-06
Create a Tekton pipeline to deploy using the GCP blueprint.

jlewi:
* This Tekton pipeline will eventually be used to continually
  deploy a fresh instance from the blueprint for CI.

Reorganize how we are defining reusable Tekton tasks.

  * Tekton tasks are currently defined in tekton/templates

  * I reorganized the tekton tasks into kustomize packages
  * I did this because I want to make it easier to hydrate the tasks
    for different installs (e.g. different namespaces).

    * e.g. for auto-deployment we will use namespace auto-deploy but
      in other settings we might use a different namespace.

Start setting up an ACM repo in acm-repo
* This will eventually be used to sync our Tekton tasks automatically
  to our cluster
* The idea is to have a single ACM repo to manage all of our CI/CD clusters
    * A single ACM repo can manage multiple clusters.
    * We could use ACM cluster selectors to select which target this applies
      to

    * So we could eventually reuse this same repo for label-sync configs
      but only sync label-sync to the cluster where label-sync runs.

* Start putting hydrated Tekton pipelines here
* ACM isn't actually installed on our cluster yet so we aren't
  actually syncing the resources yet. Right now we are still applying the manually

Update the management cluster to work for autodeployment

  * Our management cluster needs to grant kf-ci-v1-user@ GSA permissions
    to create CNRM resources so we can deploy kubeflow.
    * We do this by adding a K8s RoleBinding binding that GSA to the
      cnrm-admin ClusterRole in namespace kubeflow-ci-deployment

To support GCP blueprints I had to update the test worker image.

  * Install anthoscli, kpt, and istioctl
  * Install a newer version of yq (i.e. the yq that is a go binary
    and not a wrapper around jq).

Related to: kubeflow/gcp-blueprints#5


Bobgy:
/lgtm
I'm interested why we chose tekton pipeline. Can you point me to some previous discussion?

Issue #660, opened 2020-05-12
 Quota exceeded for quota group 'deploymentMutators' - tests are failing


Jeffwan:
We've seen some issues like below

```
Insert deployment error: googleapi: Error 403: Quota exceeded for quota group 'deploymentMutators' and limit '
Write queries per day' of service 'deploymentmanager.googleapis.com' for consumer 'project_number:29647740582'., rateLi
mitExceeded
```

I am not sure if  number of `deploymentmanager` is over the limit or write queries is over the limit. We need more investigations on this and currently tests are blocked by this issue.

/cc @jlewi

jlewi:
This is project `kubeflow-ci-deployment`

```
gcloud projects describe kubeflow-ci-deployment
createTime: '2019-02-05T01:54:33.083Z'
lifecycleState: ACTIVE
name: kubeflow-ci-deployment
parent:
  id: '1026404669954'
  type: folder
projectId: kubeflow-ci-deployment
projectNumber: '29647740582'
```

jlewi:
Here's a graph of the deployment manager API requests.
![graph](https://user-images.githubusercontent.com/777219/81754972-5617ca80-946c-11ea-90fa-9e7e533b380a.png)

The number of requests started spiking this Saturday.

I think we need to figure out what caused that spike.

Jeffwan:
> Here's a graph of the deployment manager API requests.
> ![graph](https://user-images.githubusercontent.com/777219/81754972-5617ca80-946c-11ea-90fa-9e7e533b380a.png)
> 
> The number of requests started spiking this Saturday.
> 
> I think we need to figure out what caused that spike.

It is possible to check audit logs and find caller?

Jeffwan:
I also notice one cron job failed because of `global name 'cleanup_auto_blueprints' is not defined`. Not sure if that's a problem prevent job cleaning up deploymentmanager stacks

```
+ python -m kubeflow.testing.cleanup_ci --project=kubeflow-ci-deployment --gc_backend_services=true all --delete_script=/src/kubeflow/kubeflow/scripts/gke/delete_deployment.sh
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
[...]

PR #661, opened 2020-05-13
Fix auto-deployments reconciler.

jlewi:
* The reconciler is running amok because in #657 we changed the code
  to skip over deployments with no manifest field because we couldn't get
  the zone. It turns out the lack of the manifest is an indication that
  the deployment failed. By not matching these deployments we end
  up retrying the auto-deployment. This is causing a cascading failure
  because we end up using up all our DM quota which in turn causes auto
  deployments to fail which in turn causes us to try again and eat up
  more deployments.

* This PR fixes that by matching deployments even if the manifest field
  is missing. This should prevent us from repeatedly retrying.

* Add exponential backoff to delete deployments. Because of the above
  error we are bumping to write quota limits per day which is impacting
  our ability to GC deployments.

Related to #660


Jeffwan:
looks straightforward

Issue #664, opened 2020-05-13
Recent auto-deployed clusters don't have accessible endpoints


jlewi:
The recent auto deployed (non-blueprint) deployments don't have accessible endpoints. The zone isn't set which likely indicates a problem with the deployment.

```
master | kf-master-0513-3ce | 1:33:11.410442 |   | 8061380 | kfctl-v1-0-0-g94c35cf | gcloud --project=kubeflow-ci-deployment container clusters get-credentials --zone= kf-master-0513-3ce
-- | -- | -- | -- | -- | -- | --
master | kf-master-0509-d23 | 3 days, 21:16:05.979280 |   | 928cf48 | kfctl-v1-0-0-g94c35cf | gcloud --project=kubeflow-ci-deployment container clusters get-credentials --zone= kf-master-0509-d23
stacks | kf-stacks-0513-a82 | 1:23:25.280550 |   | 8061380 | kfctl-v1-0-rc-3-35-g5b7d814 | gcloud --project=kubeflow-ci-deployment container clusters get-credentials --zone= kf-stacks-0513-a82
v1 | kf-v1-0513-ba9 | 1:33:24.761775 |   | e06c9c0 | kfctl-v1-0-0-g94c35cf | gcloud --project=kubeflow-ci-deployment container clusters get-credentials --zone= kf-v1-0513-ba9
v1 | kf-v1-0509-b55 | 3 days, 21:16:05.725653 |   | e06c9c0 | kfctl-v1-0-0-g94c35cf | gcloud --project=kubeflow-ci-deployment container clusters get-credentials --zone= kf-v1-0509-b55
```

jlewi:
Looks like the deployment failed because we hit quota issues for instance groups.

```
gcloud --project=kubeflow-ci-deployment deployment-manager deployments describe kf-v1-0513-ba9
---
fingerprint: 6l7Qty9XFpt8KguIk05w1w==
id: '2073668521133757998'
insertTime: '2020-05-13T05:32:01.505-07:00'
labels:
- key: kfctl-git
  value: kfctl-v1-0-0-g94c35cf
- key: git-manifests
  value: e06c9c0
- key: purpose
  value: kf-test-cluster
- key: git-manifests-branch
  value: v1-0-branch
- key: auto-version-name
  value: v1
- key: auto-deploy
  value: 'true'
- key: kf-name
  value: kf-v1-0513-ba9
name: kf-v1-0513-ba9
operation:
[...]

PR #665, opened 2020-05-13
Cleanup orphaned backend services and instance groups.

jlewi:
* To cleanup orphaned backend services associated with K8s ingresses
  we need to match the backend service to a URL map. If the URL map
  doesn't exist then its orphaned and we can delete it.

* Similarly for instance groups associated with K8s ingresses
  we need to map them to URL maps and delete them if the URL
  map doesn't exist.

Related to 
kubeflow/testing#660
kubeflow/testing#664

### DIFF
--- a/py/kubeflow/testing/cleanup_ci.py
+++ b/py/kubeflow/testing/cleanup_ci.py
@@ -14,7 +14,7 @@ import time
 import yaml
 
 from kubeflow.testing import argo_client
-
+from kubeflow.testing import gcp_util
 from kubeflow.testing import util
 from kubernetes import client as k8s_client
 from googleapiclient import discovery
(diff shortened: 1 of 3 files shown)


py/kubeflow/testing/cleanup_ci.py
```diff
@@ -47,6 +47,9 @@
 # resource. So we we just associae a long max age with this.
 E2E_OWNERLESS = "e2e-ownerless"
 
+DEFAULT_ZONES = ("us-east1-b,us-east1-c,us-east1-d,us-central1-a,"
```
Jeffwan, line 50: Minor: Originally it only has `"us-east1-d,us-central1-a` and it has more zones now. We may move to arguments with default in the future to avoid potential changes.

Jeffwan:
does backend services and instance group have corresponding deploymentmanagers?

jlewi:
> does backend services and instance group have corresponding deploymentmanagers?

No. These are created via Kubernetes ingress controller.

/approve

PR #672, opened 2020-05-27
Fix KFServing subarea labels

yuzisun:
The subarea labels do not show up, I think it only supports two levels.


jlewi:
I don't think that's the issue; I added the label area/kfserving/explanation to this PR. Without the  area prefix there will likely be other problems; e.g. you won't be able to add with the "/" command.

Likely something else is going. If you can maybe look at the logs for the label sync job?

Issue #682, opened 2020-06-15
tekton_client.py submits pipelineruns with names that are too long


jlewi:
See kubeflow/gcp-blueprints#50 the name ends up being

```
kubeflow-gcp-blueprints-presubmit-go-unittests-50-b1d4fcd-3488-c09d
```

This is 67 characters long.

The log of error messages looks like

```
INFO|2020-06-15T18:31:06|/src/kubeflow/testing/py/kubeflow/testing/util.py|72| name: gcp
INFO|2020-06-15T18:31:07|/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py|58| Parsing ApiException body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pipelineruns.tekton.dev \"kubeflow-gcp-blueprints-presubmit-go-unittests-50-b1d4fcd-3488-c09d\" not found","reason":"NotFound","details":{"name":"kubeflow-gcp-blueprints-presubmit-go-unittests-50-b1d4fcd-3488-c09d","group":"tekton.dev","kind":"pipelineruns"},"code":404}

INFO|2020-06-15T18:31:07|/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py|70| ApiException code=404
INFO|2020-06-15T18:31:07|/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py|87| Retry on exception: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Date': 'Mon, 15 Jun 2020 18:31:07 GMT', 'Audit-Id': 'd1512de7-c796-4d2e-a72d-d4ec4fc41b9c', 'Content-Length': '350', 'Content-Type': 'application/json'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pipelineruns.tekton.dev \"kubeflow-gcp-blueprints-presubmit-go-unittests-50-b1d4fcd-3488-c09d\" not found","reason":"NotFound","details":{"name":"kubeflow-gcp-blueprints-presubmit-go-unittests-50-b1d4fcd-3488-c09d","group":"tekton.dev","kind":"pipelineruns"},"code":404}


[...]

PR #683, opened 2020-06-15
Tekton pipeline client should use generate name

jlewi:
* Use generateName to generate names for pipelines; we can use labels
  to attach semantic/metadata information to the pipeline.

Related to #682: tekton_client.py submitting pipelineruns with names that
 are too long


jlewi:
/assign @Bobgy 

/assign @IronPan 
/assign @Ark-kun 

@Ark-kun or @IronPan would one of you mind doing a quick review of this? I need these changes to be merged to see if it fixes the issues with the Tekton pipelines. Its 4am right now in Shanghai so @Bobgy won't be around for a while.

Ark-kun:
/lgtm
There seem to be many extra changes in this PR...

jlewi:
@Ark-kun Mostly cleanup related to testing the original change

* Replacing ArgParser with Fire made it easy to add an entrypoint to allow running `tekton_client.py` locally which was one of the easiest ways to test it
* Likewise adding python3 support to run_e2e_workflow.py made it easier to run locally 
* Improved the exception handling in retries to print out the stack trace because retry logic was making it hard to see where the error was actually occuring.

PR #690, opened 2020-06-16
Update tekton.md

NikeNano:
Update path to notebook run.


NikeNano:
Found a second update in the docs.

PR #691, opened 2020-06-17
Create a task and pipeline to check IAP endpoint and KF app deployements

jlewi:
* Related to kubeflow/gcp-deployments#51; to test KF deployments are
  ready we want to create a Tekton pipeline to run the tests
  to verify if KF is ready.

* Add some dependencies to the test worker image

  * Install dateutil
  * install pytest-timeout in the worker image so we can set the --timeout

* Add a Tekton task to check if KF is ready.

  * This runs our existing test to verify that various KF applications
    are ready.

* Tasks should use the version of the kubeflow/testing/py cached in the docker image rather than checking out the testing repo.

* Related to kubeflow/gcp-deployments#52 run a test to check that
  KF applications are deployed.


jlewi:
Looks like the CLA check is having problems.

PR #704, opened 2020-06-27
GCP endpoint ready test needs to set the name of the cluster

jlewi:
* Modify get_kf_testing_cluster.py so we can output to a YAML file
  information about the cluster that we matched against.

  * This is necessary to allow getting information such as the name of
    the deployment in subsequent steps/tasks.

* Refactor get_kf_testing_cluster.py so we can start using the python Fire
  module to create CLI entrypoints as opposed to using argparse.

  * Provide backwards compatibility with argpase

Related to: kubeflow/gcp-blueprints#60 endpoint ready test is failing


jlewi:
Lint error

```
util.py                     72 INFO     E:  8, 0: No name 'get_kf_testing_cluster' in module 'kubeflow.testing' (no-name-in-module)
```

PR #707, opened 2020-06-29
Auto-deploy should use testing on master;

jlewi:
* Use ACM to deploy the auto-deploy infrastructure; previously ACM was
  only being used to deploy the Tekton tasks; we weren't including the
  other manifests in the hydration.

* Related to: kubeflow/gcp-blueprints#61


Bobgy:
/lgtm
/approve

Thanks @jlewi

PR #712, opened 2020-07-01
Fix bugs in GCP blueprint tests

jlewi:
* Related kubeflow/gcp-blueprints#51 get-credentials isn't finding any clusters
  because when using Fire the parameter should be --pattern not --location

* Related kubeflow/gcp-blueprints#65 When copying the bucket output
  in the notebook tests the parameter should be params.notebook-output
  not params.output


Bobgy:
/lgtm
Thanks

PR #714, opened 2020-07-02
Create a go binary to wrap the shell commands to create PRs.

jlewi:
To better enable GitOps we want to make it easier to create pipelines
that automatically create PRs.

This PR creates a new implementation of create_manfiests_pr.py that is
more reusable.

This PR introduces a go binary prctl that will contain subcommands
related to creating PRs.

The basic flow for automating PRs is

1. Clone the repo and create a branch on a fork repo
2. Apply some use case specific logic to modify one or more files
3. Commit and push the changes
4. Create PR from the branch

Steps 1, 3, 4 are largely the same for different use cases. With that
in mind we want to create a reusable binary so that the only part
that has to change is step 2.

There's a couple reasons for rewriting the code in go versus sticking
with Python

1. Being able to build statically linked binaries makes it a lot easier
   to create easily distrutable binaries

2. Longer term instead of shelling out to CLI tools like git and hub
   we can link and use appropriate go libraries

   * We should reach out the flux/Argo communities to see if this
     should be a part of their GitOps toolkit
     https://github.com/fluxcd/toolkit

* Related to kubeflow/kubeflow#4789
* Related to kubeflow/code-intelligence#155


jlewi:
/assign @harshad16 

@harshad16 do you want to review this?

Bobgy:
/lgtm

sounds great, KFP also want some tool to create PRs. Thank you for making a generic tool

PR #719, opened 2020-07-05
prctl can use a GitHub App to create PRs.

jlewi:
* Using GitHub apps as opposed to personal access tokens for automation
  is preferred.

  * GitHub apps are owned by the org and multiple individuals can be
    granted permission to administer them

  * This is much better than creating a shared github account

* Instead of shelling out to deprecated "hub CLI" for go; we are using
  the new GitHub CLI (github.com/cli/cli) as a library.

* Remove hub from the docker image because we no longer use it.

Related to:
kubeflow/code-intelligence#155 - continuously train and deploy label bot
kubeflow/kubeflow#4789 - Continuous builds of notebook images


Bobgy:
/lgtm
/approve

https://github.com/cli/cli
looks interesting

PR #722, opened 2020-07-08
Fix notebook test

jlewi:
* artifacts-gcs on the image build step needs to match the run-notebook
  step


jlewi:
Related to kubeflow/gcp-blueprints#65

Issue #684, opened 2020-06-15
Use python3 for run_e2e_workflow.py


jlewi:
Currently our prow jobs are firing of run_e2e_workflow.py using python2 which is no longer supported.

We should switch to using python3.

Our prow jobs are configured in:
https://github.com/kubernetes/test-infra/blob/5a71a6101fdca9159d7d6fd9e11ae8291c7fee35/config/jobs/kubeflow/kubeflow-presubmits.yaml

These basically specify a docker image to run 

```gcr.io/kubeflow-ci/test-worker:latest```

Currently this image is using the python2 image
https://github.com/kubeflow/testing/blob/master/images/Dockerfile

We do have a python3 docker image
https://github.com/kubeflow/testing/blob/master/images/Dockerfile.py3

`run_e2e_workflow.py` might need some code changes though to be python3 compatible.

scottilee:
@jlewi 
1. Which test is running `run_e2e_workflow.py`?
2. What's the process of getting a new image onto `gcr.io/kubeflow-ci/` and should it be named something like `gcr.io/kubeflow-ci/test-worker-py3:latest`?
3. Scanning through https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py I don't see anything that needs to be updated for Python 3 but how would I verify to make sure? Meaning how do I run this workflow as a test?
4. Are there any other files in that directory that you know about that need to be updated to Python 3?

jlewi:
@scottilee we should already have a python 3 image built from this dockerfile
https://github.com/kubeflow/testing/blob/master/images/Dockerfile.py3
via skaffold

These get posted to:gcr.io/kubeflow-ci/test-worker-py3
I tagged: 
gcr.io/kubeflow-ci/test-worker-py3@sha256:1b3e04157b4b27958a38b78e6f93c8741ce9a55f3bf2cbfaa50571734a6b0d06

As latest.

You can look at our Kubernetes test configs
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-presubmits.yaml

to see which configs are using run_e2e_workflow.py; pretty much all the tests should be using it. We use prow to fire off that docker container which runs run_e2e_workflow.py to fire off the E2E tests.

Maybe try picking a repository to test with and updating it to use the python3 worker image?

jlewi:
@scottilee lets start with kubeflow/testing; would you mind sending a PR my way to update the tests for kubeflow/testing?

[...]

PR #729, opened 2020-07-11
Update Dockerfile.py3

scottilee:
I added an entrypoint and grouped related things together. A few questions:

1. Comparing https://github.com/kubeflow/testing/blob/master/images/Dockerfile to https://github.com/kubeflow/testing/blob/master/images/Dockerfile.py3, there was a lot of stuff missing like Helm, Node, Yarn, Docker, and Bazel, should I add those?
2. I know Kubeflow is moving away from ksonnet, but is the project completely off of it or is it still needed for older code?
3. Can I bump up some of the versions of dependencies? For example, go1.14.2 -> go1.14.4 or hub 2.13.0 -> hub2.14.2. I'm not sure if upgrading some of these will break things.
4. Can I remove the kustomize installation? It should be included with Kubernetes automatically.
5. Once this file, Dockerfile.py3, is committed, will it automatically build and push the latest version to the registry. If not, what do I have to do to get this image pushed?

This is related to https://github.com/kubeflow/testing/issues/684 and fixing https://github.com/kubernetes/test-infra/pull/18195, which didn't seem to work as expected.

### DIFF
--- a/images/Dockerfile.py3
+++ b/images/Dockerfile.py3
@@ -1,7 +1,3 @@
-# Build the docker image used to run the scripts
-# to continuously update our docker files.
-#
-# The context for this docker file should be the root of the kubeflow/testing repository.
 FROM ubuntu:18.04
 
 RUN apt-get update -y && \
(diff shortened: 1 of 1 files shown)


jlewi:
@scottilee Lets do one thing at a time. Lets focus on fixing the tests so that we run the script to run the workflows using python3 so lets only make the changes relevant to that.

kustomize needs to be installed explicitly because I think we want a version different from what kubectl has and a lot of our scripts our explicitly calling out to kustomize not kubectl.

The image will not be built automatically. We should probably setup Cloud Build for that.

jlewi:
Please link this PR to the GitHub issue tracking using python3 as that's the motivation for this PR.

Issue #715, opened 2020-07-02
notebook tests need to use a unique subdirectory for each run in the bucket


jlewi:
The notebook tests work by submitting a job to the KF cluster that runs the notebook using papermill.
The output notebook is then uploaded as html to a GCS bucket.

Right now ever run of the test would end up using the same directory/notebook file.

Example: https://github.com/kubeflow/gcp-blueprints/issues/65#issuecomment-653047831

Looking at the logs the notebook is uploaded to

```
Uploading notebook to gs://kubeflow-ci-deployment_ci-temp/mnist_test
```

This path isn't unique for each run so we end up overwriting results.

The tricky part is that we need to set the directory consistently in two steps in the task
https://github.com/kubeflow/testing/blob/6fd81e2f38f3218b94823254e70e73f391469c01/tekton/templates/tasks/notebook-test-task.yaml#L88

1. We have the python test that fires of the job to run papermill
1. The subsequent step which copies it to a different GCS bucket

I think what we could do is could mount in the pod labels. The pod labels will include the taskrun and pipeline run labels which we know will be unique.

PR #730, opened 2020-07-11
notebook tests need to use a unique subdirectory

jlewi:
* When the notebook runs on a kubeflow-ci-deployment cluster it will
  upload the rendered notebook to a gcs bucket in kubeflow-ci-deployment.

* We need the GCS path for the notebook to be unique so that results
  from different runs don't overwrite each other.

Related to: #715


py/kubeflow/testing/notebook_tests/execute_notebook.py
```diff
@@ -62,6 +64,15 @@ def run_notebook_test(notebook_path, parameters=None):
   html_exporter = nbconvert.HTMLExporter()
   (html_output, _) = html_exporter.from_notebook_node(nb)
   gcs_path = os.getenv("OUTPUT_GCS")
+
+  # Per https://github.com/kubeflow/testing/issues/715
+  # we need to add some uniquness to the name since different test runs
+  # will use the same OUTPUT_GCS directory
+  subdir = datetime.datetime.now().strftime("%Y%m%d-%H%M")
+  subdir = "-" + uuid.uuid4().hex[0:4]
```
NikeNano, line 72: Do we wish to combine the date and the uuid? The date is just overwritten.

jlewi: @NikeNano Good catch thanks; it was a bug.

jlewi: Fixed. @NikeNano PTAL

PR #738, opened 2020-07-18
Update Dockerfile.py3 image

scottilee:
This change is to build a working Dockerfile for Python 3 (please reference: https://github.com/kubeflow/testing/issues/684). Fixes the following issues:

* .sh scripts (e.g., `checkout_repos.sh`) were not being referenced correctly causing build errors
* "py" and "notebook_testing" were being referenced although they are outside the Dockerfile directory
* hub CLI and github.com/kelseyhightower/kube-rsa were being installed twice

I was able to successfully build this Dockerfile from the "images" folder by running: `docker build -f Dockerfile.py3  .`

### DIFF
--- a/images/Dockerfile.py3
+++ b/images/Dockerfile.py3
@@ -31,13 +31,19 @@ RUN cd /tmp && \
     wget -O /tmp/go.tar.gz https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz && \
     tar -C /usr/local -xzf go.tar.gz
 
-# Install gcloud
-ENV PATH=/root/go/bin:/usr/local/go/bin:/google-cloud-sdk/bin:/workspace:${PATH} \
-    CLOUDSDK_CORE_DISABLE_PROMPTS=1
+# Create go symlinks
+RUN ln -sf /usr/local/go/bin/go /usr/local/bin && \
+    ln -sf /usr/local/go/bin/gofmt /usr/local/bin && \
+    ln -sf /usr/local/go/bin/godoc /usr/local/bin
 
 # Install the new version of yq which is based on go
 RUN GO111MODULE=on go get github.com/mikefarah/yq/v3
-RUN go get github.com/kelseyhightower/kube-rsa 
+RUN go get github.com/kelseyhightower/kube-rsa
+RUN go get -u github.com/jstemmer/go-junit-report
+
+# Install gcloud
+ENV PATH=/root/go/bin:/usr/local/go/bin:/google-cloud-sdk/bin:/workspace:${PATH} \
+    CLOUDSDK_CORE_DISABLE_PROMPTS=1
 
 RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz && \
     tar xzf google-cloud-sdk.tar.gz -C / && \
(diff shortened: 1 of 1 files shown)


scottilee:
I was able to build and run the image locally and I did receive the same error message `ModuleNotFoundError: No module named 'kubeflow'` but when I ran `export PYTHONPATH=/src/kubeflow/testing/py/` the error went away, so the `ENV PYTHONPATH /src/kubeflow/testing/py/` should fix the error but it's not... not sure what to do at this point.

images/Dockerfile.py3
```diff
@@ -71,44 +77,20 @@ RUN export ASM_VERSION=1.4.7-asm.0 && \
...
     mv kubectl /usr/local/bin && \
     chmod a+x /usr/local/bin/kubectl
 
-RUN go get -u github.com/jstemmer/go-junit-report
-
-# Create a cached copy of the python test scripts so that we don't
-# need to clone the repo just to get access to them
-RUN mkdir -p /srcCache/kubeflow/testing
-COPY py /srcCache/kubeflow/testing/py
-COPY notebook_testing /srcCache/kubeflow/testing/notebook_testing
```
jlewi, line 112: Why did you drop these commands? Some of our tests depend on a cached copy of the code.

images/Dockerfile.py3
```diff
@@ -71,44 +77,20 @@ RUN export ASM_VERSION=1.4.7-asm.0 && \
...
+ENV PYTHONPATH /src/kubeflow/testing/py/
```
jlewi, line 94: Do you need to set the python path here? Is run_workflows.sh depending on this? 
I don't think we want to set the python path because in some cases we will want to use 
/srcCache/kubeflow/testing
and sometimes we will want to use /src/kubeflow/testing

If run_workflows.sh needs the pythonpath it might be better to set the path there.

images/Dockerfile.py3
```diff
@@ -71,44 +77,20 @@ RUN export ASM_VERSION=1.4.7-asm.0 && \
...
+COPY run_release.sh /usr/local/bin/
 
-COPY ./images/run_workflows.sh /usr/local/bin
-RUN chmod a+x /usr/local/bin/run_workflows.sh
-
-COPY ./images/run_release.sh /usr/local/bin
-RUN chmod a+x /usr/local/bin/run_release.sh
-
-# Install the hub CLI for git
-RUN cd /tmp && \
```
jlewi, line 94: Why are you dropping the install of hub?

scottilee: It's duplicated.

images/Dockerfile.py3
```diff
@@ -71,44 +77,20 @@ RUN export ASM_VERSION=1.4.7-asm.0 && \
...
-    ln -sf /usr/local/go/bin/gofmt /usr/local/bin && \
-    ln -sf /usr/local/go/bin/godoc /usr/local/bin
-
-RUN go get github.com/kelseyhightower/kube-rsa
-
-COPY ./images/checkout_repos.sh /usr/local/bin
-COPY ./images/checkout.sh /usr/local/bin
-COPY ./images/setup_ssh.sh /usr/local/bin
-RUN chmod a+x /usr/local/bin/checkout* /usr/local/bin/setup_ssh.sh
+COPY checkout_repos.sh /usr/local/bin/
```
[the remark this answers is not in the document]
scottilee: Ah, when building locally it doesn't work.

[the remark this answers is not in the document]
jlewi: Did you try setting your context to be the parent directory?

images/Dockerfile.py3
jlewi: @scottilee Any thoughts on how to address this?

images/Dockerfile.py3
scottilee: I will add these back.

images/Dockerfile.py3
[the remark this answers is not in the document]
scottilee: I will try that, and will use Skaffold to test instead of just running locally. So Skaffold is a new dependency and wasn't used in the previous non-Python3 image, correct?

PR #741, opened 2020-07-22
chore: add cla yes and no labels for googlebot

Bobgy:
Part of https://github.com/kubeflow/pipelines/issues/1653

Googlebot needs "cla: yes" and "cla: no" labels in a repo before it can add/delete these labels.
We can then configure PR protection rule that targets "cla: yes" label.


Bobgy:
/approve

I think no matter what choice we want to take. It's not a problem adding these labels.

PR #759, opened 2020-10-17
Update set values tests

subodh101:
What?

As per the title.

Why?

The process of setting kfp values has been changed due to redundant variables.

### DIFF
--- a/py/kubeflow/testing/ci/kf_unittests.py
+++ b/py/kubeflow/testing/ci/kf_unittests.py
@@ -121,7 +121,7 @@ class Builder: # pylint: disable=too-many-instance-attributes
       'resources': {'limits': {'cpu': '4', 'memory': '4Gi'},
        'requests': {'cpu': '1', 'memory': '1536Mi'}},
       'volumeMounts': [{'mountPath': '/mnt/test-data-volume',
-        'name': 'kubeflow-test-volume'},
+        'name': DATA_VOLUME},
        {'mountPath': '/secret/gcp-credentials', 'name': 'gcp-credentials'}]},
      'metadata': {'labels': {
        'workflow_template': TEMPLATE_LABEL}},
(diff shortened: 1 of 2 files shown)


jlewi:
@subodh101 You should join the GitHub org so that you can trigger the tests.

subodh101:
Error:

```
mkdir -p /mnt/test-data-volume/kubeflow-testing-presubmit-py-unittests-759-147a73e-4864-5bbb/src/kubeflow
mkdir: cannot create directory ‘/mnt/test-data-volume/kubeflow-testing-presubmit-py-unittests-759-147a73e-4864-5bbb’: No space left on device
```

subodh101:
Seems like the capacity `5Gi` of the pvc `nfs-external` has been reached.

subodh101:
@Bobgy Could you help me delete some of the old directories as mentioned https://github.com/kubeflow/testing/blob/master/playbook/playbook.md#deleting-old-directories-on-the-nfs-share

Cluster: kubeflow-testing (namespace: kubeflow-test-infra)

`gcloud --project=kubeflow-ci container clusters --zone=us-east1-d get-credentials kubeflow-testing`

```
kubectl exec -it debug-worker-0 /bin/bash
Error from server (Forbidden): pods "debug-worker-0" is forbidden: User "subodh.pushkar@gmail.com" cannot create resource "pods/exec" in API group "" in the namespace "kubeflow-test-infra": requires one of ["container.pods.exec"] permission(s).
```
Also, I am having a little difficulty in checking the `disk usage`.

jlewi:
@subodh101 My preference would be to fix the issue by migrating our tests to use Tekton and eliminate the use of NFS; #769

NFS filling up has been a constant headache; so rather than doing a patch fix I think we might be better served by coming up with a more permanent solution even if it takes longer. Any chance you could pick up #769?

Bobgy:
/lgtm
Thanks for the fix!

PR #765, opened 2020-10-28
Add support to check if ingress is ready

PatrickXYS:
When developing more test cases, we found it's useful to add `wait_for_ingress ` function to check if K8s Ingress Object is ready.

### DIFF
--- a/py/kubeflow/testing/util.py
+++ b/py/kubeflow/testing/util.py
@@ -448,6 +448,49 @@ def wait_for_deployment(api_client,
     "Timeout waiting for deployment {0} in namespace {1}".format(
       name, namespace))
 
+
+def wait_for_ingress(api_client,
+                    namespace,
+                    name,
+                    timeout_minutes=2):
+  """Wait for ingress to be ready.
+
+  Args:
+    api_client: K8s api client to use.
+    namespace: The name space for the ingress.
+    name: The name of the ingress.
+    timeout_minutes: Timeout interval in minutes.
+
+  Returns:
+    ingress: The ingress object describing the ingress.
+
+  Raises:
+    TimeoutError: If timeout waiting for ingress to be ready.
+  """
+  # Wait for tiller to be ready
+  end_time = datetime.datetime.now() + datetime.timedelta(
+    minutes=timeout_minutes)
+
+  net_client_apps = k8s_client.NetworkingV1beta1Api(api_client)
+
+  while datetime.datetime.now() < end_time:
+    ingress = net_client_apps.read_namespaced_ingress(name, namespace)
+    try:
+      if len(ingress.status.load_balancer.ingress[0].hostname) == 0:
+        logging.info("Ingress %s in namespace %s is ready", name, namespace)
+        return ingress
+    except Exception:
+      logging.info("Waiting for ingress %s in namespace %s", name, namespace)
+      time.sleep(10)
+
+  logging.error("Timeout waiting for ingress %s in namespace %s to be "
+                "ready", name, namespace)
+  run(["kubectl", "describe", "ingress", "-n", namespace, name])
+  raise TimeoutError(
+    "Timeout waiting for ingress {0} in namespace {1}".format(
+      name, namespace))
+
+
 def wait_for_job(api_client,
                  namespace,
                  name,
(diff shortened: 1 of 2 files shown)


PatrickXYS:
looks like one-off error

/test kubeflow-testing-presubmit

py/kubeflow/testing/util.py
```diff
@@ -448,6 +448,48 @@ def wait_for_deployment(api_client,
...
+    ingress: The ingress object describing the ingress.
+
+  Raises:
+    TimeoutError: If timeout waiting for ingress to be ready.
+  """
+  # Wait for tiller to be ready
+  end_time = datetime.datetime.now() + datetime.timedelta(
+    minutes=timeout_minutes)
+
+  ext_client_apps = k8s_client.ExtensionsV1beta1Api(api_client)
```
Jeffwan, line 474: ingress should be `networking.k8s.io/v1beta1` in 1.16+. Can you double check API version?

py/kubeflow/testing/util.py
```diff
@@ -448,6 +448,48 @@ def wait_for_deployment(api_client,
...
+
+  while datetime.datetime.now() < end_time:
+    ingress = ext_client_apps.read_namespaced_ingress(name, namespace)
+    # ready_replicas could be None
+    if ingress.status.load_balancer.ingress[0].hostname:
```
Jeffwan, line 479: I am not sure if there's state like if ingress.status.load_balancer == false or len(if ingress.status.load_balancer.ingress) == 0. Better to double check and make sure the code here doesn't throw unexpected exception

Jeffwan:
Better to lowercase all characters beside first Char and terms.

PatrickXYS:
I'll double check APIs, maybe `try+catch` can better handle error.

py/kubeflow/testing/util.py
PatrickXYS: Checked Kubernetes 1.16 [ChangeLog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals), so ingress uses `networking.k8s.io/v1beta1` since 1.14, and `extensions/v1beta1` will be removed in v1.20

PatrickXYS: I think it make sense to use other client set to make sure it still works for v1.20+

py/kubeflow/testing/util.py
PatrickXYS: In `ExtensionsV1beta1Api ` client, there's nothing like:

```
ingress.status.load_balancer == false or len(if ingress.status.load_balancer.ingress) == 0
```

I'll check other client set

py/kubeflow/testing/util.py
```diff
@@ -448,6 +448,48 @@ def wait_for_deployment(api_client,
...
+
+def wait_for_ingress(api_client,
+                        namespace,
+                        name,
+                        timeout_minutes=2):
+  """Wait for ingress to be ready.
+
+  Args:
+    api_client: K8s api client to use.
+    namespace: The name space for the deployment.
```
andreyvelich, line 460:
```suggestion
    namespace: The name space for the ingress.
```

py/kubeflow/testing/util.py
PatrickXYS: Use `NetworkingV1beta1Api` instead

py/kubeflow/testing/util.py
PatrickXYS: Use `try + catch` to handle exception instead

py/kubeflow/testing/util.py
PatrickXYS: Good catch, thanks

PatrickXYS:
looks like one-off error

/test kubeflow-testing-presubmit

PatrickXYS:
```
ERROR|2020-10-29T17:01:07|/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py|46| KeyError: 'status'
  File "/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py", line 41, in log_status
    workflow["status"]["conditions"][0]["reason"])
KeyError: 'status'
INFO|2020-10-29T17:01:07|/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py|88| Retry on exception: 'status'; stack trace:
  File "/usr/local/lib/python2.7/dist-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/src/kubeflow/testing/py/kubeflow/testing/tekton_client.py", line 119, in get_namespaced_custom_object_with_retries
    condition = result["status"]["conditions"][0]["reason"]
KeyError: 'status'
```
Looks like error related to Tekton


/test kubeflow-testing-presubmit

Jeffwan (approved):

PatrickXYS:
Known Issue to TektonPipeline, tracked here 

https://github.com/kubeflow/gcp-blueprints/issues/143

Issue #750, opened 2020-08-25
GC old tekton resources periodically


Bobgy:
Discovered in https://github.com/kubeflow/gcp-blueprints/issues/128.

Currently all triggered `PipelineRun` and other tekton resources accumulate for ever.
We need to GC tekton resources.

Tekton community has an issue tektoncd/experimental#479 tracking this.
But there's no ready-to-use solution yet.

Before sth comes out of above issue, we should have at least a bash script that cleans up these resources.

Bobgy:
I wrote a very simple bash script for this purpose:
```bash
#!/bin/bash

set -e

# This script deletes k8s resources in batches of 100.
# usage: ./rm_k8s_res.sh <input-file-path>
# for example: ./rm_k8s_res.sh ./input.txt
# get the input file by
# ```
# kubectl get <resource-type> --sort-by=.metadata.creationTimestamp -o name > input.txt
# ```
# input file is a list of resource names to be deleted ordered by creation timestamp.
# if only want to delete old resources, you can manually remove the last section to keep latest resources.

input="$1"
resource="$2"
echo "Deleting $2 from file $input"
names=()
while IFS=" " read -r name age
do
  names+=($name)
  if [ ${#names[@]} -gt 100 ]; then
    echo "Deleting ${names[@]}"
    kubectl delete ${names[@]} || echo "already deleted"
    names=()
  fi
done < "$input"
if [ ${#names[@]} -gt 0 ]; then
  echo "Deleting ${names[@]}"
  kubectl delete ${names[@]} || echo "already deleted"
fi
```

Bobgy:
After deleting to ~100 pipelineruns left in the cluster, https://kf-ci-v1.endpoints.kubeflow-ci.cloud.goog/tekton/#/namespaces/auto-deploy/pipelineruns can successfully load pipelineruns now.

jlewi:
@Bobgy FYI this is the script we use to cleanup argo workflows.
https://github.com/kubeflow/testing/blob/d4394eaff73af5645d156412f1de374aba31bf33/py/kubeflow/testing/cleanup_ci.py#L112

jlewi:
#767 has a very simple program. The program can be used with arbitrary K8s resources and can delete things based on minimum age.

We need to turn that into a regularly running job with a suitable minimum age (e.g. 72 hours).

We can use the same pattern as we do for groups sync:
https://github.com/kubeflow/internal-acls/tree/master/google_groups/manifests

[...]

PR #767, opened 2020-10-29
A simple program to GC old resources like Tekton pipelineruns

jlewi:
* The accumulation of finished jobs appears to cause performance
  issues in our test clusters

* in kubeflow/gcp-blueprints#143 we observe ~ 49K pipelineruns

* `kubectl delete -n kf-ci pipelineruns --all` was not wondering

   * It looks like the connection to the master was interuppted during a
     stream

* This program takes advantage of the List method's support for pagination

* This program also makes it possible to set a minimum age for a resource
  to be eligible for deletion.

* Related to kubeflow/gcp-blueprints#143

* Related to #750


subodh101:
I had one doubt.
Earlier I was getting the pod name from the following command.

```
kubectl get PipelineRun deploy-kf-master-txrwf -o yaml
```

Then using that pod name from the status of the PipelineRun I got the logs from gcp logging.

Just incase, is it now possible to get the pod name after 3 days somehow?

go/cmd/cleanup/cleanup_pipelineruns.yaml
```diff
@@ -0,0 +1,13 @@
...
+  groups:
+    - Group: "tekton.dev"
+      Version: "v1alpha1"
+      Namespace: "kf-ci"
+      MinAge: "72h"
+      Resource: "pipelineruns"
+    - Group: "tekton.dev"
+      Version: "v1alpha1"
+      Namespace: "kf-ci"
+      MinAge: "72h"
```
subodh101, line 12: If the above is difficult. Let's set the MinAge to "168h" i.e. one week. Because anyways the pipeline runs have been there from ages.

jlewi: You can identify pods using labels. The pod's labels include the pipelineRun as well as taskRun etc... Pod logs are preserved in stackdriver and in stackdriver you can search by labels.

```
  labels:
    app.kubernetes.io/managed-by: configmanagement.gke.io
    pipeline: xgboost-notebook
    tekton.dev/pipeline: notebook-test
    tekton.dev/pipelineRun: xgboost-vqsjr
    tekton.dev/pipelineTask: build-image
    tekton.dev/task: notebook-test-builder
    tekton.dev/taskRun: xgboost-vqsjr-build-image-4ppqn

```

Bobgy: +1 for storing for a week initially, it will still be easier debugging from pipelinerun resource.

Bobgy:
/approve

Thank you! Needs this tool for a long time.
It is generally useful in other cases.

go/cmd/cleanup/cleanup_pipelineruns.yaml
jlewi: @Bobgy Agree. Lets set it appropriately when we automate it per #750 
right now it was pretty aggressive because it was a one time cleanup.

jlewi:
I believe test failure is because NFS is out of space; #769

jlewi:
@Bobgy you'll need to relgtm because the tests were fixed.

PR #777, opened 2020-11-05
test: fix cleanup_ci.py and set up cronjob

Bobgy:
Fixes https://github.com/kubeflow/gcp-blueprints/issues/145
/assign @jlewi @chensun 

Changes:
* Fixed the cleanup_ci.py script to clean up `deleted:xxxxx` iam policy bindings
* Added a cronjob in `kubeflow-ci-v1/auto-deploy`to run this script periodically


Bobgy:
@jlewi Did we set up periodic job for cleanup_ci.py any where?
I feel it's very important to reduce test infra flakiness.

jlewi:
@Bobgy We should have a cron job.
https://github.com/kubeflow/testing/blob/master/test-infra/cleanup/cleanup-ci-cronjob.yaml

I think when we checked on Thursday the cron job was in an error state saying it had missed too many executions.

Bobgy:
Getting this in to unblock https://github.com/kubeflow/gcp-blueprints/issues/145

PR #775, opened 2020-11-04
Revert Code Change

PatrickXYS:
Resolve #774 

/cc @jlewi @Jeffwan  

This effects all the kubeflow repos which have not yet migrated to Tekton

Issue #769, opened 2020-10-30
Convert Python unittest Argo Workflow to Tekton


jlewi:
Per #768 we need to get rid of the dependency on NFS by converting our Argo workflow for running python unittests to using Tekton.

The Argo Workflow is defined here.
https://github.com/kubeflow/testing/blob/4e02994179cfbdc3f3dd93ffb7c6ded6c9aad11c/py/kubeflow/testing/ci/kf_unittests.py#L192

We can model our Tekton pipeline off of our go unittests pipeline
https://github.com/kubeflow/testing/blob/master/tekton/runs/go-test-run.yaml
https://github.com/kubeflow/testing/blob/master/tekton/templates/pipelines/go-test-pipeline.yaml
https://github.com/kubeflow/testing/blob/master/tekton/templates/tasks/go-tests.yaml

The overall structure is largely the same

* We need to create two Tekton tasks
  
  * The first will run lint; i.e. this [argo step](https://github.com/kubeflow/testing/blob/4e02994179cfbdc3f3dd93ffb7c6ded6c9aad11c/py/kubeflow/testing/ci/kf_unittests.py#L217)
  * The second will run the unittests; i.e. this [argo step](https://github.com/kubeflow/testing/blob/4e02994179cfbdc3f3dd93ffb7c6ded6c9aad11c/py/kubeflow/testing/ci/kf_unittests.py#L197)

These tasks will be very similar to 
https://github.com/kubeflow/testing/blob/master/tekton/templates/tasks/go-tests.yaml

i.e. there will be two steps in each
* Run the tests or lint
* Run copy-artifacts

/cc @subodh101 @Jeffwan @PatrickXYS

subodh101:
@PatrickXYS I think I would be able to do it.
Also, I have started working on it in case of double work. Sorry for not assigning earlier.

PatrickXYS:
@subodh101 Sounds good, thanks for helping!

Jeffwan:
@subodh101 @PatrickXYS 

When you help on the migration, please try not to delete Argo related codes unless you are 100% sure it's not being used. 
https://github.com/kubeflow/testing/issues/774

We can have some clean up later once 1.2 get released. We have a few PRs blocked this morning

jlewi:
> When you help on the migration, please try not to delete Argo related codes unless you are 100% sure it's not being used.
#774

The deleted code was ksonnet code. I was the one who said to delete it.

[...]

PR #778, opened 2020-11-05
Create individual pipeline from pipelinerun to make it reusable

subodh101:
What?
Reference: https://github.com/kubeflow/testing/issues/769
- [x] Separate py test pipeline from run https://github.com/kubeflow/testing/issues/769#issuecomment-722390032.
- [x] Add python path as per the https://github.com/kubeflow/testing/issues/769#issuecomment-722407594.
- ~Add lint in the name as per the https://github.com/kubeflow/testing/pull/775#issuecomment-721947828.~


subodh101:
mostly it is same.
I have just added lint in the name.
Tested it and confirms it works.

PatrickXYS:
@subodh101 Thanks! I'm pretty sure we need pylint our code since we enabled py-unittest, so my question is what step does user need to do before sending PR and received error before not pylint their code?

Can we document somewhere or change the PR template say something like

```
make sure you finish these steps before sending PR

[All the steps that are required]
```

subodh101:
> @subodh101 Thanks! I'm pretty sure we need pylint our code since we enabled py-unittest, so my question is what step does user need to do before sending PR and received error if not pylint their code?
> 
> Can we document somewhere or change the PR template say something like
> 
> ```
> make sure you finish these steps before sending PR
> 
> - pytest py/kubeflow/testing/py_lint_test.py -s --rcfile=.pylintrc
> ```

Oh. thank you. Yes, that sounds good. Let's do it in a separate PR.
I think we need to create an idea issues PR for what can be done to do faster development.

tekton/templates/pipelines/py-unittests-pipeline.yaml
```diff
@@ -0,0 +1,36 @@
...
+    params:
+    - name: artifacts-gcs
+      value: "$(params.artifacts-gcs)"
+    resources:
+      inputs:
+      - name: testing-repo
+        resource: testing-repo
+    taskRef:
+      name: py-test
+    runAfter:
```
jlewi, line 35: Delete this. The unittests and lint can run in parallel which should be faster.

jlewi:
Thanks for the quick fix.

subodh101:
@jlewi It is running. Looks good now! Please merge.

PR #782, opened 2020-11-06
Update management ACM config

jlewi:
* ACM config should be sync'ing to kubeflow/testing@master


Bobgy:
/lgtm
/approve
Do I need to manually deploy the change after merge?

PR #780, opened 2020-11-06
Run updated tekton pipelinerun py-unittests once the pipeline is created

subodh101:
What?
As per the title.
Waiting on https://github.com/kubeflow/testing/pull/778 to go in.

PR #794, opened 2020-11-08
Update py unittest

subodh101:
What?
- add source path for lint and unit test.
- Update to py docker image.
- Fix python path for all the steps.

Usage:
- path for the lint and unit test directory is to be set at the pipelinerun based on the locations in any repository.
For eg. in case of kubeflow/testing repo:

```
  - name: lint-src-dir
    value: "py"
  - name: test-src-dir
    value: "py/kubeflow/tests"
``` 
- .pylintrc file should be present at the toplevel of any repository using it.

Ref: https://github.com/kubeflow/testing/issues/769

### DIFF
--- a/acm-repos/kf-ci-v1/namespaces/auto-deploy/tekton.dev_v1alpha1_pipeline_py-unittests.yaml
+++ b/acm-repos/kf-ci-v1/namespaces/auto-deploy/tekton.dev_v1alpha1_pipeline_py-unittests.yaml
@@ -9,6 +9,17 @@ spec:
       in the form of 'gs://'
     name: artifacts-gcs
     type: string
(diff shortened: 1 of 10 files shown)


Bobgy (commented):
/lgtm
Thank you @subodh101!

/assign @jlewi 
for a second review

subodh101:
/assign @jlewi
PTAL, 
the test is commented out at the moment.
So others may be merging their PR's without running py-unittest.

Which can cause a little trouble in merging:
https://github.com/kubeflow/testing/pull/780

jlewi:
Thanks. Apologies for the slow reply. I was on vacation.

/lgtm
/approve

PR #804, opened 2020-11-10
Create v1.2.0 config

Jeffwan:
I think this is the configuration for CD. I cut v1.2-branch in following repos. Once it's added, they should build images with v1.2 tag and update kubeflow/manifest repo. It also helps on cherry-picks against v1.2-branch. 

/cc @PatrickXYS @jlewi @Bobgy @pingsutw

### DIFF
--- a/apps-cd/applications.yaml
+++ b/apps-cd/applications.yaml
@@ -265,3 +265,24 @@ versions:
               value: v1.1-branch
             - name: url
               value: git@github.com:kubeflow/tf-operator.git
+  # Define a v1-2 release
+  - name: v1-2
+    # A tag to prefix image names with
+    tag: "v1.2.0"
+    repos:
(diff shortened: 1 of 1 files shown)


PatrickXYS:
/approve

>I think this is the configuration for CD. I cut v1.2-branch in following repos. Once it's added, they should build images with v1.2 tag and update kubeflow/manifest repo. It also helps on cherry-picks against v1.2-branch.

I think the expected behavior should be like above, @jlewi can help check if possible

Bobgy:
/lgtm
/approve
Thanks @Jeffwan

Jeffwan:
@Bobgy @jlewi   Seems CD is down. I see some images built on Sept. I am not sure if all the latest changes have corresponding images now.. This might be a blocking issue.

/cc @thesuperzapper

PR #808, opened 2020-11-14
Add a workload identity binding for the emojichat namespace

jlewi:
* Needed in order to download models from GCS.


jlewi:
/assign @Bobgy 

I don't know why the bot doesn't think the tests should run. I checked and it looks like I'm a member of the Kubeflow org.

jlewi:
I take that back looks like I'm a member of the Kubeflow incubator org but not Kubeflow.

jlewi:
I had a new invitation that I needed to accept; I'm guessing I was removed as part of downgrading me from a GitHub org admin.

jlewi:
/approve

@Bobgy force submitting this to unblock myself.

PR #816, opened 2020-11-20
test(gcp): revert set values commands for kubeflow package

Bobgy:
Depended on by https://github.com/kubeflow/gcp-blueprints/pull/169


subodh101 (approved):
/lgtm
Thanks

PR #817, opened 2020-11-21
Add reusable scripts to kubeflow/testing

Jeffwan:
Move some helpful scripts to kubeflow/testing.

See https://github.com/kubeflow/kubeflow/issues/5419 for more details


PatrickXYS:
This is great! @Jeffwan 

/lgtm

scripts/update-changelog.sh
```diff
@@ -0,0 +1,44 @@
...
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." > /dev/null && pwd)"
+
+cd ${ROOT}
+
+if [ "${GITHUB_TOKEN}" == "NO" ]; then
+  echo "Environment variable GITHUB_TOKEN is not set."
+  exit 1
+fi
+
+github_changelog_generator -t ${GITHUB_TOKEN} -u kubeflow -p kubeflow \
```
Bobgy, line 38: Is there any repo still using this?
KFP was using this tool, but once the repo size grow to a certain limit, it always fail on github rate limit.

If we do not find existing users, I'd suggest getting rid of it.

Bobgy (commented):
one minor comment for confirmation

thanks for moving these!

scripts/update-changelog.sh
Jeffwan: I don't know any users on this now.  

For 1.2 release on kuebflow/kfctl/manifest. I use a different log generator scripts and manually categorize PRs. 

I am fine to remove it.  What do you use in KFP now? If there's a recommended way, we should remove it since it will encounter rate limit issue

Jeffwan:
Update: Remove github_changelog_generator as per Bobgy's suggestion. If

Bobgy:
You may find my investigations in https://github.com/kubeflow/pipelines/issues/3920. KFP now uses a subset of https://www.conventionalcommits.org/en/v1.0.0/ conventions to generate the changelog.

Bobgy:
Thanks for the update!
/lgtm
/approve

PR #819, opened 2020-11-24
test: fix notebook_runner dockerfile

Bobgy:
Part of https://github.com/kubeflow/gcp-blueprints/issues/174


Bobgy:
self LGTM to verify in upstream

I verified the fix allowed the dockerfile to build.

PR #825, opened 2020-11-27
test(gcp): refactor standar cluster package with instance and upstream, adjust node locations

Bobgy:
/cc @chensun 
second PR following https://github.com/kubeflow/testing/pull/824


Bobgy:
Going to merge to take effect on the cluster

PR #828, opened 2020-11-28
test(kfp): use staging proxy instead

Bobgy:
/cc @chensun


Bobgy:
/hold
we'll need this project added to inverse proxy staging env first

PR #836, opened 2020-12-16
Add Doc and GitOps resources for Optional-Test-Infra

PatrickXYS:
The PR content is described as below:

1. Documentation for Optiona-Test-Infra
2. GitOps to manipulate Optional-Test-Infra clusters, related to https://github.com/kubeflow/testing/issues/791

Very initial PR to improve transparency of optional-test-infra, and should not impact any other cloud-providers

PR #837, opened 2020-12-17
Fix bad file name

PatrickXYS:
Fix bad file name introduced from init PR https://github.com/kubeflow/testing/pull/836


PatrickXYS:
Merge fast to unblock, have no effect on existing testing-infra

PR #865, opened 2021-01-23
Update optional-test-infra doc

PatrickXYS:
Resolves #

1. Update optional-test-infra documentation
2. Make root README clean up, and divide documentation into oss-test-infra and optional-test-infra parts.

**Checklist:**

- [ ] Changes have been generated to `aws/GitOps` folder: 
    1. `cd aws`
    2. `make optional-generate`
    3. `make optional-test`


README.md
```diff
@@ -1,1110 +1,8 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
PatrickXYS, line 1: @Bobgy I cleaned up the README and divided this into two parts, oss-test-infra and optional-test-infra.

Let me know if you want us to better organize the documentation

images/Dockerfile.py3.aws
```diff
@@ -14,9 +14,6 @@ RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py
 RUN python3.8 -m pip install \
     filelock \
     fire \
-    google-api-python-client \
```
PatrickXYS, line 17: I think this is unnecessary to have those dependencies in optional-test-infra worker image.

README.md
Jeffwan: project root level does need a README.md. I would suggest to move some content out of the root README.md instead of deleting it

PatrickXYS: I didn't delete those doc but moved it into gcp_README.md. 

Maybe I can add some common content back to README.md, what do you think?

Jeffwan: sounds good. Seems some contents are general information here.

PatrickXYS: Moved contents back to root README

PatrickXYS:
/cc @Bobgy @Jeffwan 
Can you take another look?

PatrickXYS:
I'll merge it as of now, if we have any other concerns, we can open another PR to update doc

PR #901, opened 2021-02-16
Kaniko cache

theofpa:
Add the caching option in kaniko task to have the build tasks running faster

**Checklist:**

- [x] Changes need to be generated to `aws/GitOps` folder: 
    1. `cd aws`
    2. `make optional-generate`
    3. `make optional-test`

### DIFF
--- a/aws/GitOps/clusters/optional-test-infra-tekton/namespaces/kubeflow-test-infra/kaniko.yaml
+++ b/aws/GitOps/clusters/optional-test-infra-tekton/namespaces/kubeflow-test-infra/kaniko.yaml
@@ -23,6 +23,8 @@ spec:
   - name: SUBPATH
     description: Sub path within the given context.
     default: ""
+  - name: REVISION
+    description: Revision (commit SHA1) of the git repository to clone and the docker repository to tag.
   - name: EXTRA_ARGS
     default: ""
   - name: BUILDER_IMAGE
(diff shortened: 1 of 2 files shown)


PatrickXYS:
I'm not sure if this will make task running faster, trying to understand how it works in Tekton.

> Set this flag as --cache=true to opt into caching with kaniko.

Reference from Kaniko page, if the parameter enables the functionality of caching in local files, note: we're running on the cloud instead of the local machine, how can we take advantage of caching instead? Can you help elaborate a bit?

In general, I think the PR's intention is good, I'm not sure if this takes into effect in our test-infra

theofpa:
I'm planning to use it for the kfserving project. We build the following images:

| Image                | Image size | Build&push time |
|----------------------|------------|-----------------|
| alibi-explainer      | 2.48GB     | 7m34s           |
| kfserving-controller | 22MB       | 9m36s           |
| pytorchserver-gpu    | 1.63GB     | 18m23s          |
| pytorchserver        | 660MB      | 6m7s            |
| sklearnserver        | 165MB      | 1m39s           |
| storage-initializer  | 98MB       | 38s             |
| xgbserver            | 400MB      | 2m14s           |
| aix-explainer        | 5.2GB      | 8m2s            |
| pmmlserver           | 523MB      | 1m59s           |
| custom-transformer   | 1.82GB     | 5m57s           |
| art-explainer        | 698MB      | 4m7s            |
| lgbserver            | 289MB      | 1m46s           |
| agent                | 829MB      | 9m36s           |

Most of these images do not change on every PR, and it’s a pity to spend cpu cycles to build them again and again.

Kaniko supports caching of the layers which have not been changed, by using the remote repository, ECR-public in our case, addressed in 769b8d0.

A problem seen with the remote layers caching is that due to the large size of some layers (ex. Alibi-explainer, aix-explainer, custom-transformer), the build step fails with OMMKilled error. 

The cpu/memory requests can be configured per step in the kaniko task definition, we need to check if this can be configured in the task of these particular images which use the kaniko taskref. To have a successful push in the cache of these images, I had to increase the memory of the build step to 14GB, which is not acceptable. 

We need to check the kaniko warmer image and how it works.

PatrickXYS:
@theofpa Can you dry-run commands again? 

```
cd aws
make optional-generate
make optional-test
```

PatrickXYS:
Meanwhile let's keep investing into Kaniko and monitor our testing issue.

I think it would be a good point to contribute with our actual experience.

/lgtm
/approve 
/hold 

Hold until test succeed

PatrickXYS (approved):
/lgtm
/approve
/unhold

PR #915, opened 2021-02-26
Ported for PPC64LE

ezeeyahoo:
Hi,

Tested on both PPC64LE and X86_64, works well for me. 

Please have a look on queries below and review.
cc: @gerrith3

### DIFF
--- a/images/Dockerfile.py3_ppc
+++ b/images/Dockerfile.py3_ppc
@@ -1,0 +1,121 @@
+# Build the docker image used to run the scripts
+# to continuously update our docker files.
+#
+# The context for this docker file should be the root of the kubeflow/testing repository.
+FROM ubuntu:18.04
+
+ARG ARCH=amd64
+ARG GITHUB_PROFILE=ezeeyahoo
+
+RUN apt-get update -y --fix-missing && \
+    apt-get install -y curl git python3.8 python3-distutils wget && \
+    ln -sf /usr/bin/python3.8 /usr/bin/python && \
+    ln -sf /usr/bin/python3.8 /usr/bin/python3
+
+RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py
+RUN if [ "$ARCH" = "ppc64le" ] ; then apt-get install build-essential \
+    libssl-dev libffi-dev python3.8-dev -y; fi
+RUN python3.8 -m pip install \
+    filelock \
+    fire \
+    google-api-python-client \
+    google-cloud \
+    google-cloud-storage \
+    junit-xml \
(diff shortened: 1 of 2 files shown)


PatrickXYS:
I know the PR is still WIP, just a heads up, does anyone use PPC64LE architecture for now... 

Not sure if this is needed...

gerrith3:
@PatrickXYS YES - ppc64le is used to run many of the world's top commerce, banking, retail, transportation, etc. workloads.  It is not near as visible/common on the desktop but under the covers ppc64le is very common and the number of requests we have been getting for a working kubeflow has been increasing.  Kubeflow is an optional component that can be added to various solutions that we are working on, and a number of clients are asking for support directly as well.

PatrickXYS:
Let me make it clear, I'm not against PPC64LE architecture, it's great we have kubeflow onboard the architecture, but why we have to do in testing image? Or more specifically, who will run presubmit/postsubmit/periodic with the testing image of PPC64LE? 

Again, very welcome to contribute, all the questions proposed here is to help you avoid useless efforts.

gerrith3:
@PatrickXYS Patrick, first, thanks for the dialog!  We definitely appreciate the ability to compare notes and ensure we are delivering this "correctly".  As to your question about why we would self-host a repo - simply because we don't include kubeflow as it stands in a product.  We do that for a number of open source components, especially where we want to lock down versions and changes.  However, we don't currently include kubeflow in a product.  Instead, several solutions/products tell customers/end users that (on Intel) they should get the packages "from the community" and trust that the community is doing proper hygiene, addressing CVEs, packaging and testing releases, etc.  The challenge here is that many of those products are build for Intel, Power and sometimes even our Z brand - and the existing documentation says "if you want to use kubeflow with this product, get kubeflow from the community" - which doesn't work if the community code is not ported to Power or current.

As far as pre-submit/postsubmit and even periodic, we have made Travis CI available for Power testing, and it should be very easy to ensure that all pull requests are tested on travis on ppc64le prior to acceptance, and it is also possible to schedule release builds with travis.  @ezeeyahoo can help set that up with you if necessary.

BTW, @ezeeyahoo is working to sign the CLA - since it is a stock Linux Foundation CLA that shouldn't be an issue, we have an internal process to get people listed and Eshant is started on that path now and that should be resolved in the next couple of days.

Let me know if you have other questions!

PatrickXYS:
@gerrith3 I got your point

> we have made Travis CI available for Power testing, and it should be very easy to ensure that all pull requests are tested on travis on ppc64le prior to acceptance

If this is the case, there's no need to make this PR, the testing image build based on the dockerfile is used for Prow, or more specifically Kubeflow community's prow. 

I'd suggest you looking into documentation to make sure this align with your goal

PR #968, opened 2021-11-01
test(kfp): upgrade to 1.7.1

zijianjoy:


Bobgy:
Did you run make hydrate-kfp-manifests?
It generates raw k8s resources in acm-repos folder which is source of truth for the cluster via gitops.

zijianjoy:
@Bobgy thank you for the info! I just ran this command and added information in README.

Bobgy:
/lgtm
Thank you!

PR #986, opened 2022-02-09
test(kfp): Remove status from kfp-standalone-1 manifest

chensun:
Resolves #

**Checklist:**

    1. `cd aws`
    2. `make optional-generate`
    3. `make optional-test`

### DIFF
--- a/acm-repos/kfp-standalone-1/kfp-all.yaml
+++ b/acm-repos/kfp-standalone-1/kfp-all.yaml
@@ -531,13 +531,6 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  $patch: delete
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition


acm-repos/kfp-standalone-1/kfp-all.yaml
```diff
@@ -531,13 +531,6 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  $patch: delete
```
chensun, line 535: @zijianjoy This patch doesn't seem to work: https://github.com/kubeflow/testing/blob/3d41370cbc4027ddadfd9a5b73cf389dfc247fc0/test-infra/kfp/kfp-standalone-1/kustomize/instance/status-deletion.yaml

I did some quick search , but haven't found the correct way of doing it yet.

zijianjoy: What is your kustomize version?

Mine is 
{Version:kustomize/v4.4.0 GitCommit:63ec6bdb3d737a7c66901828c5743656c49b60e1 BuildDate:2021-09-27T16:24:12Z GoOs:darwin GoArch:amd64}

Reference of me running the `make kfp-update` command: https://github.com/kubeflow/testing/pull/987

chensun: Okay, mine was 
Version: {Version:kustomize/v3.2.1 GitCommit:d89b448c745937f0cf1936162f26a5aac688f840 BuildDate:2019-09-27T00:10:52Z GoOs:linux GoArch:amd64}
That's probably why, will try a newer version later.

chensun: @zijianjoy, actually after I upgrade my kustomize to 
{Version:kustomize/v4.5.2 GitCommit:9091919699baf1c5a5bf71b32ca73a993e98088b BuildDate:2022-02-09T23:26:42Z GoOs:linux GoArch:amd64}
I still get the same result.

zijianjoy: Hmm, are you using the `kubectl kustomize` or `kustomize`? You might have multiple kustomize installed.

Otherwise, it is difficult for me to debug as I cannot reproduce. You can also try to change the patch https://github.com/kubeflow/testing/blob/master/test-infra/kfp/kfp-standalone-1/kustomize/instance/kustomization.yaml#L34 to use something differently. For example: patchesJSON6902 (And you need to change the status-delete.yaml to a different format in json). For more patch approaches, see: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/inlinePatch.md#inline-patch-for-patchesjson6902

chensun: > are you using the `kubectl kustomize` or `kustomize`?

@zijianjoy That's a good point. I'm using the Makefile, so it's `kubectl kustomize`. Not sure what's the difference or relationship between the two, but after I upgrade my `kubectl` to latest, it works for me now. Thanks!

PR #989, opened 2022-02-11
chore(kfp): update instructions for updating kfp deployment in the test cluster.

chensun:
Resolves #

`hydrate-kfp-manifests` is already covered by `kfp-update`, so no need to run that separately again.

https://github.com/kubeflow/testing/blob/5a610d067789d63eb5771884946387702d497e57/test-infra/kfp/Makefile#L54

**Checklist:**

    1. `cd aws`
    2. `make optional-generate`
    3. `make optional-test`

### DIFF
--- a/test-infra/kfp/README.md
+++ b/test-infra/kfp/README.md
@@ -2,6 +2,14 @@
 
 ## Upgrade KFP
 
+1. (Optionally) Run:
+
+    ```bash
+    make hydrate-kfp-manifests
+    ```
+    
+    To check the generated raw k8s resources in acm-repos folder based on local changes (without pulling manifests from kfp repo). 
+
 1. Edit `PIPELINES_VERSION=<new-version>` in Makefile.
 
 1. Run:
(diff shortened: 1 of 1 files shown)


test-infra/kfp/README.md
```diff
@@ -10,12 +10,6 @@
     make kfp-update
     ```
 
-1. Run:
-
-    ```bash
-    make hydrate-kfp-manifests
-    ```
-
     It generates raw k8s resources in acm-repos folder which is source of truth for the cluster via gitops.
```
zijianjoy, line 13: Can we keep this instruction? Instead of removing it, maybe we should give more guidance to users about how to use it.

From my opinion: We should run `make hydrate-kfp-manifests` first. It is important for editor to be able to review the generated manifests, before pushing it to the CI cluster. After validation, editor can run `make kfp-update` to apply the change.

chensun: Updated. PTAL.

zijianjoy:
/lgtm
/approve

Thank you Chen!

PR #1013, opened 2022-08-24
kfp: Reduce persistent agent TTL_SECONDS_AFTER_WORKFLOW_FINISH to 8 hours

chensun:
Reduce persistent agent TTL_SECONDS_AFTER_WORKFLOW_FINISH from 24 hours to 8 hours.

`kubeflow-pipelines-samples-v2` are often flaky and deleting completed pods--usually a lot--seems to reduce the flakiness. We can probably reduce the default TTL to 8 hours as I expect most test debugging should happen within such time window.

### DIFF
--- a/acm-repos/kfp-standalone-1/kfp-all.yaml
+++ b/acm-repos/kfp-standalone-1/kfp-all.yaml
@@ -3118,12 +3118,12 @@ spec:
     spec:
       containers:
       - env:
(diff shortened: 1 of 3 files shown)


connor-mccarthy (approved):
/lgtm
/approve

Good idea! Thanks for adding this.

PR #1021, opened 2022-12-21
Updates to fix bug on latest godoc extraction and fail on error

ukclivecox:
- Fixes godoc extraction of repo
- Fails if repo can not be found


christian-heusel:
/close
Closing since the underlying code is no longer present, see https://github.com/kubeflow/testing/pull/1065 for details.

PR #1067, opened 2026-08-05
Add juliusvonkohout as approver

juliusvonkohout:
@andreyvelich lets see what i can upstream from the community distribution tests.


juliusvonkohout:
The tests have to be refactored, i will force merge.

andreyvelich:
Thanks for your help @juliusvonkohout!
Let's ensure we only push common scripts and actions that are used across multiple subprojects, like boilerplate check:
https://github.com/kubeflow/testing/tree/master/hack/boilerplate

As we discussed with @thesuperzapper that subprojects will continue to maintain their specific tests.

PR #1072, opened 2026-08-27
feat: add reusable create-cluster (Kind) composite action

TrueFurina:
## Summary

Adds a reusable composite action (`.github/actions/create-cluster`) that spins up a local Kubernetes cluster (Kind) so CI jobs can run against a real cluster instead of each workflow re-implementing the setup inline. This mirrors the pattern already used by `sdk` and `trainer` e2e workflows.

## Requirements met

- Composite action with inputs for **Kubernetes version** and **Kind node image** (both with sensible defaults)
- Exports `KUBECONFIG` (via `$GITHUB_ENV`) so subsequent steps can talk to the cluster directly
- Optional cluster name input

## Changes

- New file: `.github/actions/create-cluster/action.yml`

## Verification

- YAML follows the composite-action spec (`runs.using: composite`)
- Inputs match the issue requirements


.github/actions/create-cluster/action.yml
```diff
@@ -0,0 +1,36 @@
+name: "Create Kind Cluster"
+description: >-
+  Spins up a local Kubernetes cluster (Kind) so CI jobs can run against a real
+  cluster without each workflow re-implementing the setup inline. Exports
+  KUBECONFIG so subsequent steps can talk to the cluster directly.
+inputs:
+  kubernetes-version:
+    description: Kubernetes version for the cluster.
+    required: false
+    default: v1.31.0
```
juliusvonkohout, line 10: 1.36 please take a look at the workflow in the community distribution  as reference

juliusvonkohout: Actually we will soon upgrade to 1.37 https://github.com/kubernetes-sigs/kind/releases/tag/v0.33.0

tariq-hasan:
Hi @TrueFurina! Checking in to see if you are still working on the PR. There are some pending comments above that needed to be addressed.

TrueFurina:
Thanks @juliusvonkohout — updated the composite action to Kubernetes 1.37 (kind v0.33.0):

- Bumped `helm/kind-action` to `v1.15.0`.
- Pinned `kind-node-image` to `kindest/node:v1.37.0@sha256:a1ed56cfb0e7b93589bdf97c8cd566405a265939e3620fc4f5de89adff580ae5` (the v0.33.0 default node image).
- Fixed the version mapping: `version` is now the Kind CLI version `v0.33.0` (it was previously mistakenly passed the Kubernetes version), and `kubectl_version` is driven by the `kubernetes-version` input.
- Added the referenced `kind-config.yaml` (single control-plane node) — the action previously pointed at a config file that wasn't committed, which would have failed at runtime.

Let me know if you'd prefer a different node-image pin or a multi-node config.

TrueFurina:
Addressed all three points from the Copilot review, plus the version bump requested earlier.

1. **Removed the self-checkout step.** Agreed — a composite action shouldn't reset a workspace it doesn't own. Callers have already checked out the repository, and re-checking could discard a different ref or generated files before the caller's next steps. The action now relies on the caller's checkout.

2. **Pinned `helm/kind-action` to an immutable commit** — `06c1ae10762d3b9c1644e7fe69596ae519e015a2`, with `# v1.15.0` retained as a comment, matching the convention in `.github/workflows/github-stale.yaml`.

3. **`KUBECONFIG` now preserves the caller's value.** The export step was unconditionally writing `$HOME/.kube/config`, so if a job had already exported `KUBECONFIG` (and Kind honours it unless an explicit `--kubeconfig` is passed), the cluster would be written to one path while later steps read another. It now re-exports the existing value when set and only falls back to the default location when unset.

Also carried over from @juliusvonkohout's earlier comments: Kubernetes **1.37** via kind **v0.33.0**, with `kind-config.yaml` added (it was referenced but missing, which would have failed at runtime).

Head is `022c861f`. Happy to adjust the KUBECONFIG handling if you'd rather pass one explicit path through to Kind instead.

juliusvonkohout:
Please check the failing dco

